PR #3, opened 2018-07-07
WIP: Initial version of the json report structure schema

legopiraat:
Initial setup for the json schema.

I've added a field for line number as well. I think that was missing in the initial json. I think `lineNumber` in combination with `span` is needed for highlighting correct?

`mutatorName` could maybe be renamed to `mutatorCategory` as it's more a category of the mutation as they are not named individually. 

Also think that `language` might be able to be moved to the root of the json. Not sure if `stryker-net` and `stryker` are able to mutate multiple languages in the same run but `stryker4s` only handles `scala` so we do not need a `language` per mutator.

TODOs

- [ ] Define which files are required
- [ ] Add meaning descriptions on the fields with TODO in them.

### DIFF
--- a/mutation-testing-report-schema.json
+++ b/mutation-testing-report-schema.json
@@ -1,0 +1,216 @@
+{
+    "$id": "http://example.com/example.json",
+    "type": "object",
+    "definitions": {},
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "properties": {
+      "name": {
+        "$id": "/properties/name",
+        "type": "string",
+        "title": "TODO",
+        "default": "",
+        "examples": [
+          "src"
+        ]
+      },
+      "path": {
+        "$id": "/properties/path",
+        "type": "string",
+        "title": "Absolute path to the mutated source code.",
+        "default": "",
+        "examples": [
+          "/usr/full/path/to/src"
+        ]
+      },
+      "totals": {
+        "$id": "/properties/totals",
+        "type": "object",
+        "properties": {
+          "any": {
+            "$id": "/properties/totals/properties/any",
+            "type": "integer",
+            "title": "Key value pairs with mutation information like: mutationsScore, killed, survived etc.",
+            "default": 0,
+            "examples": [
+              8
+            ]
+          }
+        }
+      },
+      "health": {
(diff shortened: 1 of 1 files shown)


nicojs:
>  I think `lineNumber` in combination with `span` is needed for highlighting correct?

With `span` I meant something like `range`. I.e. "the amount of characters since the beginning of the file".

For example, take this file:

```
12345
67890
```

In this example:
* number 1 is at span `[0,1]`.
* numbers 2 and 3 are at span `[2,4]`
* number 6 is at span `[6,7]` on linux/mac, but `[7,8]` on windows (allowing for carriage return).

This illustrates nicely what the problem is. We should either: 

* Align on one kind of line endings (`\n` would be the one to go for); or
* Use line numbers in combination with columns

Using line numbers is slightly more hassle for the web component itself, but negligible. What do you think?

> `mutatorName` could maybe be renamed to `mutatorCategory` as it's more a category of the mutation as they are not named individually.

Well... that depends. You could argue that an IncrementMutator (name "increment") could belong to a category "BinaryMutators", which in turn could also belong to a broader category "OperatorMutators". 

I would reserve the word `Category` for something that we can add later. For now, keep it flat with use a mutator name. We can decide to allign on names, but don't have to (and it is not possible for all languages).

> Also think that `language` might be able to be moved to the root of the json. 

Stryker already supports both JavaScript and Typescript. I think we should keep it as it is. We could however choose to remove it and let the reporter decide which language it is based on extension.

legopiraat:
Some tests would be great, i probably won't have to time to complete tests before my vacation so if you could implement them that would be great. 

On the ` lineNumbers:column` vs `span` discussion, maybe we should discuss what would be easier during the hackathon maybe and keep it as span for now.

legopiraat:
I've added two small test cases. One with a valid json file and one which misses a property. If you want to extend (or completely change it) these tests be my guest.

hugo-vrijswijk:
Is the idea to add the HTML reporter as a git submodule to each project, or to make it available through some other means? As a dependency it would work better with versioning, of course, but that would be a lot of work for each language. Having the JSON schema available for each project would also help with integration into the project

PR #10, opened 2019-02-24
Feat html elements

nicojs:


packages/mutation-testing-elements/src/components/mutation-test-report-totals.ts
```diff
@@ -0,0 +1,135 @@
...
+      <div><span># Timeout</span></div>
+    </th>
+    <th class="rotate text-center" style="width: 50px">
+      <div><span># No coverage</span></div>
+    </th>
+    <th class="rotate text-center" style="width: 50px">
+      <div><span># Runtime errors</span></div>
+    </th>
+    <th class="rotate text-center" style="width: 50px">
+      <div><span># Transpile errors</span></div>
```
Mobrockers: The property is called CompileError in the schema

nicojs: Thanks! I will change that.

nicojs:
I finally think the quality is good enough :D ready for review. I think we can release version 1 after merge.

Note, this is functionally the same as v 0.0.7 (with actual tests 😅 ).

nicojs:
image

Hope you guys agree 👍

Mobrockers:
We have already integrated 0.7.0 so lgtm.

nicojs:
Thanks @Mobrockers. If you've got the time, a code review is welcome as well.

packages/mutation-testing-elements/src/typings/highlight.js.d.ts
```diff
@@ -0,0 +1,27 @@
...
+declare module 'highlight.js/lib/languages/javascript' {
+  export = Language;
+}
+declare module 'highlight.js/lib/languages/typescript' {
+  export = Language;
+}
+declare module 'highlight.js/lib/languages/scala' {
+  export = Language;
+}
+declare module 'highlight.js/lib/languages/java' {
```
hugo-vrijswijk: Why is Java imported? As there are no frameworks using this report that have Java support. I feel like we should either import all Highlight.js languages, or add them on a case by case basis (preferred)

packages/mutation-testing-elements/src/components/mutation-test-report-file.ts
```diff
@@ -0,0 +1,95 @@
...
+@customElement('mutation-test-report-file')
+export class MutationTestReportFileComponent extends LitElement {
+
+  @property()
+  public model!: FileResultModel;
+
+  public static styles = [
+    highlightJS,
+    bootstrap,
+    css`
```
hugo-vrijswijk: Shouldn't styles like this be in a separate css (or sass) file, so that Webpack can properly optimize them (and, if we would want, use magic like autoprefixer and such with it)

packages/mutation-testing-elements/src/components/mutation-test-report-totals.ts
```diff
@@ -0,0 +1,185 @@
...
+    </tbody>`;
+  }
+
+  private renderRow(name: string, row: ResultModel, shouldLink: boolean) {
+    const mutationScoreRounded = row.totals.mutationScore.toFixed(2);
+    const coloringClass = this.determineColoringClass(row.totals.mutationScore);
+    const progressBarStyle = `width: ${mutationScoreRounded}%`;
+    return html`
+    <tr title="${row.name}">
+      <td style="width: 17px;" class="icon no-border-right">${row.representsFile ? svg.file : svg.directory}</td>
```
hugo-vrijswijk: I find the look of this element to be a little weird (see image). The `<` looks like it's clickable, but it's not. And because it is a `<` it also looks like a back button, even though it is for going 'deeper' into the directory structure

image

packages/mutation-testing-elements/src/typings/highlight.js.d.ts
nicojs: True. Removed 👍

packages/mutation-testing-elements/src/components/mutation-test-report-file.ts
nicojs: Good catch! Both `bootstrap` and `highlightjs` are separated into different files.  We should do the same for custom files.

packages/mutation-testing-elements/src/components/mutation-test-report-totals.ts
nicojs: I don't see the `<`:

image

Can you send me a deep link?

nicojs:
@hugo-vrijswijk  I've got the feeling that the report renders differently on your side. I don't see the `<` or the misalignments.

packages/mutation-testing-elements/src/components/mutation-test-report-file.ts
```diff
@@ -0,0 +1,94 @@
+import { LitElement, html, property, customElement, css, PropertyValues } from 'lit-element';
+import { unsafeHTML } from 'lit-html/directives/unsafe-html';
+import hljs from 'highlight.js/lib/highlight';
```
hugo-vrijswijk: Adding to it, in my opinion the default highlight.js theme is quite ugly. Perhaps we could use a different one. [There are quite a few nice ones](https://highlightjs.org/static/demo/). On stryker-mutator.io we use the GitHub theme. Atom One Light is also nice. Or Monokai or Atom One Dark would be my suggestions if we want to have a dark theme

packages/mutation-testing-elements/src/components/mutation-test-report-totals.ts
hugo-vrijswijk: Weird. I just ran the report with `npm start` and clicked through it. Maybe there's a difference between dev and prod?

hugo-vrijswijk: Ok I am not sure what I was doing yesterday, but I don't have the issue anymore

nicojs: I think you typed the `>` yourself 😅

packages/mutation-testing-elements/src/components/mutation-test-report-file.ts
nicojs: > we use the GitHub theme. Atom One Light is also nice. Or Monokai or Atom One Dark would be my suggestions if we want to have a dark theme

So GitHub? 😅

hugo-vrijswijk: I mean, if you want to. I personally prefer some of the others though. Choosing the theme shouldn't be based on what's already on the website, as that can also be changed easily

nicojs: I'll add "github" for now. We can later make it dynamic by adding it as a property (`<mutation-test-report-app highlight-theme="github"></mutation-test-report-app>`)

nicojs:
I think all review comments have been addressed. OK to merge?

hugo-vrijswijk:
🚢🚀

PR #13, opened 2019-03-11
Remove `additionalProperties: false`, and add regex to schemaVersion

hugo-vrijswijk:
The `additionalProperties: false` might conflict with future additions to the schema and backwards compatibility. For example, a new property being added in 1.1, would not be a valid schema for 1.0.

Also added a regex to the schema version. Starts with 1, and then optionally a dot and any number.


nicojs:
> The `additionalProperties: false` might conflict with future additions to the schema and backward compatibility

Thanks a lot! I've been thinking that we should use this schema to validate any report-json pushes to the dashboard in the future, to makes sure people don't upload arbitrary JSON/JS (prevent any change of javascript injection).

PR #17, opened 2019-03-24
feat(mutant popup): add popup for selected mutant

### LINKED ISSUE #16: Where to host additional information like "mutatorName"
Where do we want to host additional information about the mutators? Like MutatorName? This is now missing entirely from the report, but it might be useful. For example, you can disable certain mutators by name, just as soon as you know the mutator name.

In the old report, we had a popup using bootstrap js:

image

I don't want to load the entire bootstrap js (it isn't "webpackable" and relies on JQuery 😢 ). So I would need to recreate it. At the same time, I'm not really happy with putting this info in a popup.

We also had a table with all the info:

image

I wasn't happy with that either. It can feel like an overload of info. It also took valuable space at the right side of the screen, resulting in a horizontal scrollbar for the code area.

What solution do we want? 

We can do something like a drawer panel and fit the table in that. We could make it so that you can activate, or deactivate a mutant from there as well. https://www.w3schools.com/howto/tryit.asp?filename=tryhow_js_sidenav

Other ideas?

nicojs:
This PR adds a popup when selecting a mutant. For example:

image

The styling is based on bootstrap. The placement is based on https://www.w3schools.com/howto/howto_js_popup.asp

Fixes #16


package.json
```diff
@@ -59,12 +59,14 @@
...
     "mocha": "^6.0.2",
     "node-sass": "^4.11.0",
+    "popper.js": "^1.14.7",
     "rimraf": "^2.6.3",
     "sass-loader": "^7.1.0",
     "selenium-webdriver": "^4.0.0-alpha.1",
     "sinon": "^7.2.7",
     "source-map-support": "^0.5.10",
     "terser-webpack-plugin": "^1.2.3",
+    "tooltip.js": "^1.3.1",
```
nicojs: Should be removed

package.json
```diff
@@ -59,12 +59,14 @@
     "lit-element": "^2.0.1",
     "mocha": "^6.0.2",
     "node-sass": "^4.11.0",
+    "popper.js": "^1.14.7",
```
nicojs: Should be removed

packages/mutation-testing-elements/src/components/mutation-test-report-popup/mutation-test-report-popup.scss
```diff
@@ -0,0 +1,35 @@
...
+
+/* The actual popup */
+.popover {
+    width: 200px;
+    position: absolute;
+    z-index: 1;
+    top: 125%;
+    left: 50%;
+    margin-left: -100px;
+    // visibility: hidden;
```
nicojs: This should be enabled! Right now, a popup with opacity 1 can be blocking a mutant!

PR #21, opened 2019-04-04
feat(sonatype): Add Sonatype release

hugo-vrijswijk:
Adds a release to sonatype. Works similar to the [Stryker4s maven plugin](https://github.com/stryker-mutator/stryker4s/tree/master/runners/maven). The maven project is empty, except for  A script is run that sets the version of the project to the one set in the `package.json`, the signing secret is imported, and `mvn deploy` is called, which will create a jar containing only the `dist` folder, and upload that to Sonatype.

I've also added four secrets to the travis build


hugo-vrijswijk:
> Could you add a remark to the readme on how to use it with as a maven dependency?
Will do!

> How will we deploy the schema to maven? Would that need the same kind of setup? Should that be added in this PR?
Yeah that should be a similar setup. I'll add that to this PR as well.

I'll also move the packaged resources to a sub-folder to prevent any conflicts

hugo-vrijswijk:
So the current setup is not ideal. It will now build a maven release for every project, instead of only for the projects that Lerna sees are changed and builds. I'm not sure how to change the build script for Lerna to also include the maven project

hugo-vrijswijk:
Ready for review again! The build now uses the `postpack` step to import gpg key data in the root project, and `postpublish` in each project to run `mvn deploy`

packages/mutation-testing-report-schema/mvn/pom.xml
```diff
@@ -0,0 +1,110 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.stryker-mutator</groupId>
+    <artifactId>mutation-testing-report-schema</artifactId>
+    <version>1.0.0</version>
```
nicojs:
```suggestion
     <version>SET-BY-SCRIPT-SNAPSHOT</version>
```

PR #22, opened 2019-04-04
Visual changes/fixes

hugo-vrijswijk:
I've made some visual changes and fixes. It's easiest (I think) to review per commit, as that shows what changes as for what component. I just felt like separate PR's would be a bit much

The changes:
- The status text in the popup no longer has different paddings top and bottom
- The little arrow on the popup is now the same colour as the rest of the popover
Before:
image
After:
image
- The legend has some padding and is properly on the foreground
- The header no longer overlaps with the file breadcrumb
Before (header overlap and no padding):
image
Before (foreground issues):
image
After (header):
image
After (sticky header when scrolling down):
image
- The whole filename column is now clickable instead of just the filename (hard to show in a screenshot)


nicojs:
Love the branch name. Look 👀! Changed 🔀!

packages/mutation-testing-elements/src/components/mutation-test-report-popup/mutation-test-report-popup.scss
```diff
@@ -18,18 +18,20 @@
 }
 
 /* Popup arrow */
-.popover::before {
+.popover-arrow {
```
nicojs: I rather use a `::before` pseudo-element instead of an actual div with classes

packages/mutation-testing-elements/src/components/mutation-test-report-popup/mutation-test-report-popup.ts
```diff
@@ -28,17 +28,22 @@ export class MutationTestReportPopupComponent extends LitElement {
...
-    if (this.context) {
-      return `bg-${this.context} text-white`;
-    } else {
-      return '';
-    }
+  private getContextClasses() {
+      return this.getIfContext(`bg-${this.context} text-white`);
+  }
+
+  private getBorderClasses() {
```
nicojs: See my previous remark.

packages/mutation-testing-elements/src/components/mutation-test-report-popup/mutation-test-report-popup.scss
hugo-vrijswijk: The issue I ran into was that to dynamically set the colour based on the context, the `border-danger` (or other) from bootstrap is used. That class is set from ts code, so the element needs to be accessible from the ts code. For `::before` I couldn't find a way to do that

packages/mutation-testing-elements/src/components/mutation-test-report-popup/mutation-test-report-popup.scss
```diff
@@ -18,18 +18,20 @@
...
     position: absolute;
     top: -10px;
     left: 50%;
-    margin-left: -5px;
-    border-width: 5px;
+    margin-left: -6px;
+    border-width: 6px;
     border-style: solid;
-    border-color: transparent transparent #555 transparent;
+    border-left-color: transparent !important;
```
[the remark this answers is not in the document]
hugo-vrijswijk: The border-color is set by the bootstrap class (see below comment), which makes the whole element that colour. To override it and only make the arrow (border-bottom) the colour I used `!important`. I'm open to suggestions on other ways to do it, because it does feel a bit hacky (but most things I do in CSS do)

packages/mutation-testing-elements/src/components/mutation-test-report-popup/mutation-test-report-popup.scss
nicojs: I did find a way, see new implementation 👍

packages/mutation-testing-elements/src/components/mutation-test-report-popup/mutation-test-report-popup.scss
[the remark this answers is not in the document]
nicojs: Fixed with my new implementation 👍

nicojs:
Also included muted title postfix (slightly grey and smaller): 

image

packages/mutation-testing-elements/src/components/mutation-test-report-popup/mutation-test-report-popup.scss
hugo-vrijswijk: Very nice! Does that mean the borders import from bootstrap can also be removed?

nicojs: Yep, we just did 👍

nicojs:
I've reworded the commits, will merge when build is green

PR #28, opened 2019-04-06
feat(metrics): move metrics to it's own package

nicojs:
Move the calculation of the metrics to it's own package.
This way we can reuse it in the console reporter and the
stryker dashboard.


nicojs:
Our integration tests are a thing of beauty 😎. I've got high confidence even when changing around a lot like in this PR.

hugo-vrijswijk:
If I understand correctly, the plan is to have multiple language implementations for `mutation-testing-metrics` (which I think is a great idea, to have some sort of standard). Shouldn't the npm package be under a subfolder in `packages/mutation-testing-metrics`? So a `node`, `jvm`, `.net`, etc folder?

nicojs:
We can add that once we want to create it, right? No need to have that now. The npm package content would remain the same.

PR #32, opened 2019-04-18
feat(TypeScript): publish type information

simondel:


packages/mutation-testing-metrics/.npmignore
```diff
@@ -1,3 +1,4 @@
 **/*
 !README.md
 !dist/src/**/*.js
+!dist/src/**/*.js.map
```
nicojs: Should be `*.d.ts`

simondel: Prrrt! Fixed

PR #35, opened 2019-04-19
build: add build for node 8

simondel:


nicojs:
Why is this needed? Mutation-testing-elements runs in the browser, not node, right?

simondel:
Good point. Kill it!

nicojs:
It's a bit harsh to kill mutation-testing-elements just because it runs a browser. I mean, it's not his fault. 🤔

hugo-vrijswijk:
Alternatively, metrics could run in node

nicojs:
Yeah good point.

PR #38, opened 2019-05-10
Use dart sass instead of node-sass implementation

hugo-vrijswijk:


nicojs:
Awesome! Never break node-bindings for sass again 😅

Issue #18, opened 2019-04-03
Support PIT - the Java mutation testing framework


nicojs:
[PIT](https://github.com/hcoles/pitest) is interested in using mutation testing elements  see https://twitter.com/0hjc/status/1113369365066473473

We need to support mutants that don't have an exact location or replacement. Only line number and mutator name.

Example JSON:

```json
{ 
  "schemaVersion": "1.1",
   "thresholds": {
    "high": 80,
    "low": 60
  },
    "files": {
       "src/main/foo.java": {
           "language": "java",
           "source": "public class Progam {\n    public static void main(String[] args) {\n    System.out.println(\"bar\");\n}",
          "mutants": [
            {
              "id": "1",
              "mutatorName": "StringLiteral",
              "location": {
                "start": {
                  "line": 3
                }
              },
              "status": "Survived"
            }
          ]
       }
```

@hcoles 
Is this json something PIT could produce? Also: mutation-testing-elements is opinionated: `"mutatorName"` is a free text field, but `"status"` isn't. Do you agree with our definitions and metrics? Documented here: https://github.com/stryker-mutator/stryker-handbook/blob/master/mutant-states-and-metrics.md

hugo-vrijswijk:
Requiring a replacement when a column is present, but not otherwise might be able to be enforced in the schema by using [property dependencies](https://json-schema.org/understanding-json-schema/reference/object.html#dependencies)

hugo-vrijswijk:
This might also be a good time to look at a 'real' release to maven central, instead of using [webjars](http://webjars.org/)

nicojs:
I would separate the two issues. Feel free to add an automated release to maven 💪.

hcoles:
@nicojs 

The status definitions match very closely to what PIT does (even the names match). We have the additional (rare) status of "memory error" but this can be mapped to runtime error.

As PIT can't provide a replacement it provides a description of the mutation in addition to the mutator name. For some mutators this can provide useful contextual information, for others it is a fixed description.

The mutator name and description could perhaps be concatenated together to fit this schema, but from my point of view it would be preferable to keep them separate if possible.

hcoles:
I've created an issue on the pitest side. 

https://github.com/hcoles/pitest/issues/592

Unfortunately it will need to wait on finding some-one with more time than I have to look at it.

nicojs:
[...]

PR #43, opened 2019-06-06
Feat/emoj and no coverage color

nicojs:
This PR adds a new orange color to differentiate no-coverage from survived states more easily. This should make the report more in line with what PIT does with its line coverage. 

It also adds emojis to the mutant legends. Both are demoed here: 

image

### DIFF
--- a/packages/mutation-testing-elements/src/lib/htmlHelpers.ts
+++ b/packages/mutation-testing-elements/src/lib/htmlHelpers.ts
@@ -5,6 +5,7 @@ export function getContextClassForStatus(status: MutantStatus) {
     case MutantStatus.Killed:
       return 'success';
     case MutantStatus.NoCoverage:
+      return 'caution'; // custom class
     case MutantStatus.Survived:
       return 'danger';
     case MutantStatus.Timeout:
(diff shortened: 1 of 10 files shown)


nicojs:
It's not necessary to squash and merge this one. Rebase and merge is preferred (for changelog purposes).

nicojs:
@hcoles do you agree that #18 can be closed with this new addition (especially the color, but maybe you fancy the emojis as well 😍 ).

packages/mutation-testing-elements/test/unit/components/mutation-test-report-file-legend.spec.ts
```diff
@@ -34,16 +35,40 @@ describe(MutationTestReportFileLegendComponent.name, () => {
...
+      '✅ Killed (1)',
+      '👽 Survived (1)',
+      '🙈 NoCoverage (1)',
+      '⌛ Timeout (1)',
+      '💥 CompileError (1)',
+      '💥 RuntimeError (1)'
     ]);
   });
 
+  it('should show the correct colors for the states', async () => {
```
hugo-vrijswijk: No idea if this is possible in JS, but wouldn't it be better to do something like:
```js
expectedMutantColors.forEach((status, expectedColor) => {
  it(`should show the color ${expectedColor} for status ${status}`, () => ...);
});
```

Edit: great minds think alike? You already have that in a test below. So why not here?

packages/mutation-testing-elements/src/lib/htmlHelpers.ts
```diff
@@ -18,15 +19,16 @@ export function getContextClassForStatus(status: MutantStatus) {
 export function getEmojiForStatus(status: MutantStatus) {
   switch (status) {
     case MutantStatus.Killed:
-      return '✔';
+      return '✅';
     case MutantStatus.NoCoverage:
+      return '🙈';
     case MutantStatus.Survived:
-      return '❌';
+      return '👽';
```
hugo-vrijswijk: 🎉

packages/mutation-testing-elements/test/unit/components/mutation-test-report-file-legend.spec.ts
nicojs: With "below", you mean in another file 😌

No idea why I guess I wrote them on separate days? Fixing it now...

PR #45, opened 2019-07-06
fix(legend): status alignments

hugo-vrijswijk:
Before:
image

After:
image

Legends are now center-aligned. Huge difference, I know. But it annoyed me.


nicojs:
I'll trust that there is a difference here.

PR #52, opened 2019-07-12
feat(Nullable mutation score): Mutation score nullable to support 0 mutations in a file

richardwerkman:
Make the mutation score nullable to support 0 mutations in a file

image

Print as `N/A` instead of 100%


packages/mutation-testing-elements/package-lock.json
```diff
@@ -24,6 +24,22 @@
       "resolved": "https://registry.npmjs.org/lodash.groupby/-/lodash.groupby-4.6.0.tgz",
       "integrity": "sha1-Cwih3PaDl8OXhVwyOXg4Mt90A9E=",
       "dev": true
+    },
+    "mutation-testing-metrics": {
```
nicojs: I think you ran `npm install` directly in `packages/mutation-testing-elements`. Please run `npm install` (which in turn runs `lerna bootstrap`) in the root of the project to fix this.

packages/mutation-testing-elements/src/components/mutation-test-report-totals.ts
```diff
@@ -143,7 +143,12 @@ export class MutationTestReportTotalsComponent extends LitElement {
   }
 
   private renderRow(name: string, row: MetricsResult, path: string | undefined) {
-    const mutationScoreRounded = row.metrics.mutationScore.toFixed(2);
+    let mutationScore = 0;
+    let mutationScoreRounded = null;
+    if (row.metrics.mutationScore != null) {
+      mutationScore = row.metrics.mutationScore as number;
```
nicojs: This `as number` cast shouldn't be needed. `row.metrics.mutationScore != null` is a type guard.

packages/mutation-testing-elements/src/components/mutation-test-report-totals.ts
```diff
@@ -155,11 +160,11 @@ export class MutationTestReportTotalsComponent extends LitElement {
         <div class="progress">
           <div class="progress-bar bg-${coloringClass}" role="progressbar" aria-valuenow="${mutationScoreRounded}"
             aria-valuemin="0" aria-valuemax="100" .style="${progressBarStyle}">
-            ${mutationScoreRounded}%
+            ${mutationScoreRounded ? mutationScoreRounded + '%' : ''}
```
nicojs: Could we add the text `'Not available'` in the progress bar?

packages/mutation-testing-elements/src/components/mutation-test-report-totals.ts
```diff
@@ -143,7 +143,12 @@ export class MutationTestReportTotalsComponent extends LitElement {
   }
 
   private renderRow(name: string, row: MetricsResult, path: string | undefined) {
-    const mutationScoreRounded = row.metrics.mutationScore.toFixed(2);
+    let mutationScore = 0;
+    let mutationScoreRounded = null;
```
nicojs: Please create a helper function to calculate `mutationScoreRounded`, instead of resorting to `let`.

packages/mutation-testing-elements/src/components/mutation-test-report-totals.ts
```diff
@@ -155,11 +160,11 @@ export class MutationTestReportTotalsComponent extends LitElement {
         <div class="progress">
           <div class="progress-bar bg-${coloringClass}" role="progressbar" aria-valuenow="${mutationScoreRounded}"
             aria-valuemin="0" aria-valuemax="100" .style="${progressBarStyle}">
-            ${mutationScoreRounded}%
+            ${mutationScoreRounded ? mutationScoreRounded + '%' : ''}
           </div>
         </div>
       </td>
-      <th style="width: 50px;" class="no-border-left text-center text-${coloringClass}">${mutationScoreRounded}</th>
+      <th style="width: 50px;" class="no-border-left text-center text-${coloringClass}">${mutationScoreRounded ? mutationScoreRounded : 'N/A'}</th>
```
nicojs: What happens when `mutationScore` is `0`? That's a falsy value, so probably will be `n/a`. Shouldn't be, right?

hugo-vrijswijk:
@richardwerkman How would you feel about using NaN for a score when there are no mutants, and checking on NaN in the report instead?

packages/mutation-testing-elements/src/components/mutation-test-report-totals/index.ts
```diff
@@ -128,11 +130,18 @@ export class MutationTestReportTotalsComponent extends LitElement {
     </tr>`;
   }
 
-  private determineColoringClass(score: number) {
-    if (this.thresholds) {
-      if (score < this.thresholds.low) {
+  private mutationScoreRounded(mutationScore: number) {
+    if (!isNaN(mutationScore)) {
```
hugo-vrijswijk: This if is not needed anymore:

```sh
> 3.3.toFixed(2)
'3.30'
> NaN.toFixed(2)
'NaN'
```

richardwerkman: And then later I should check for the string 'NaN'?

packages/mutation-testing-metrics/src/api/Metrics.ts
```diff
@@ -59,12 +59,12 @@ export interface Metrics {
    */
   totalCovered: number;
   /**
-   * The total percentage of mutants that were killed.
+   * The total percentage of mutants that were killed, or null if not available (n/a).
```
hugo-vrijswijk:
```suggestion
   * The total percentage of mutants that were killed, or NaN if not available.
```

packages/mutation-testing-metrics/src/api/Metrics.ts
```diff
@@ -59,12 +59,12 @@ export interface Metrics {
...
    * `totalDetected / totalValid * 100`,
    */
   mutationScore: number;
   /**
-   * The total percentage of mutants that were killed based on the code coverage results of the initial test run.
+   * The total percentage of mutants that were killed based on the code coverage results of the initial test run, or null if not available (n/a).
```
hugo-vrijswijk:
```suggestion
   * The total percentage of mutants that were killed based on the code coverage results of the initial test run, or NaN if not available.
```

hugo-vrijswijk:
@richardwerkman I've changed the code to be a little more DRY, as there were a lot of NaN checks, I hope you don't mind. The blue 'Not available' text also looked a little weird to me. It now looks like this:
image

Let me know what you think, feel free to revert my commit if you disagree!

nicojs:
Probably better to not show an empty progress bar in case of `N/A`, right?

hugo-vrijswijk:
Should there be no progress bar at all?

nicojs:
Nope. An empty progress bar is reserved for mutation score `0%` 😎

nicojs:
Merging this after green build. Good job both!

PR #65, opened 2019-08-23
feat: add mutation-testing-metrics for Scala

hugo-vrijswijk:
Adds a mutation testing metrics for Scala. Cross-compiles to Scala 2.11, 2.12, 2.13, Dotty, ScalaJS & ScalaNative.

Up for discussion: the package name. Right now it fallls under `io.stryker.mutator.metrics`. Is this what we want? It connects it more to the Stryker name. I am open to suggestions.

TODO:
- [x] Actual implementation
- [x] Testing
- [x] Readme documentation
- [x] Figure out publishing process:
  - [x] Configure Travis build
  - [x] Push artifacts to Sonatype
  - [x] Integrate into Lerna process:
    - [x] testing
    - [x] publishing
    - [x] setting the correct version (read from package.json in same directory)


nicojs:
Cool! I like this as it could be reused by any jvm language, right? Maybe we should remove `scala` from the name, use jvm instead? 

About the maven coordinates. Is it possible to not use `io.stryker.mutator`? Instead something with `mutation-testing`?

hugo-vrijswijk:
That's a good idea! Depending on the implementation, though. My primary goal was to provide a good experience for Scala users. I've never used Scala classes from Java before, so I wonder how well it works (case classes, objects, etc). Something to look at, I guess.

As for the coordinates, I think using `io.stryker-mutator` in the groupId name is sort of necessary. We don't own any other domain to publish under (and mutation-testing.io is a bit expensive, I just checked). That's also what the `elements` mirror is published under. The package name could simply be `mutation.testing` though (no dashes).

nicojs:
> I've never used Scala classes from Java before, so I wonder how well it works (case classes, objects, etc). Something to look at, I guess.

Should be pretty straightforward. I think it is fine to focus on the scala use case and still call it `jvm`. If other API's are needed, people can add them.

> As for the coordinates, I think using io.stryker-mutator in the groupId name is sort of necessary

Let's use the stryker-mutator coordinate with mutationtesting package.

hugo-vrijswijk:
@nicojs Thinking about the name, `mutation-testing-metrics-jvm` also does not cover the whole story. ScalaJS and Scala Native don't run on the JVM. How about `mutation-testing-metrics-scala-211-212-213-dotty-and-scalajs-and-scala-native-but-probably-also-jvm`? With naming conventions like these I could get a job at Microsoft

Wmaarts:
> Thinking about the name, mutation-testing-metrics-jvm also does not cover the whole story. 

Just calling it `scala` could work too, since it's primarily scala as a use case. And it's no secret that Scala's cross compatible. Could also make it clear somewhere else (thinking of `README`) that this supports Java too, once that's verified. :)

hugo-vrijswijk:
This PR is now ready for review. I've added some extra tests to verify the metrics against the JSON schema. 

Publishing via Travis 'should' work. It's a little hard to test, but I've plugged into Lerna lifecycle commands for the sbt project (and that works for tests).

@nicojs @Wmaarts @legopiraat Could one of you have a look for a review?

PR #70, opened 2019-08-25
build(deps-dev): update chromedriver

hugo-vrijswijk:
I noticed that the build was failing on basically every branch. This is because of an old chromedriver not working with a newer version of Chrome (76). This updates the chromedriver to work with it.


nicojs:
Awesome, thx!

Issue #25, opened 2019-04-05
Project icon


richardwerkman:
Right now only files and folders are seperated. It would be nice if we could also mark a folder as a project. 

![image](https://user-images.githubusercontent.com/14224500/55626490-f7354b80-57ab-11e9-9d1d-88bbffc149b0.png)

Here we have two projects: `SharedProject` and `TestApp`. Right now it could be confusing to see the projects as folders. Maybe a different icon could help?

nicojs:
So use the directory icon here? Might make more sense indeed...

I don't feel much for introducing  "project" icon. We would need some kind of knowledge of what is a project and what are files/directories.

hugo-vrijswijk:
I don't think we should use the directory icon here. It click through to a file, so it makes sense to be a file

nicojs:
It would be cool to add the "official" file type icons, as vscode does 😍 
![image](https://user-images.githubusercontent.com/1828233/55627713-27cab480-57af-11e9-8f15-236ad0f571e1.png)

But that takes too much time for now (and little return of value 💰 )

PR #173, opened 2019-10-28
feat(icons): Add icons to file a few types

### LINKED ISSUE #26: File type icons
It'd be cool to have file type specific icons, similar to VS Code. We could use something like [pretty-file-icons](https://www.npmjs.com/package/pretty-file-icons) for it. Mapping icons by hand is probably not a very good idea.

Note: `pretty-file-icons` seems to be specific for https://www.flaticon.com, which I am not a huge fan of

image

legopiraat:
@nicojs @hugo-vrijswijk 
This is an initial setup to add more specific icons to file types based on the file extension.

I would like some input if this is something we want. 
It's easy to add a missing file type from
https://github.com/vscode-icons/vscode-icons/tree/master/icons

Adding an icon can be done by defining the SVG and adding it to the mapping. It's retrieved based on file extensions. 

This will fix #26  and #25


legopiraat:
I am going to rewrite how the icons are injected since I would like to remove the dependency I added it doesn't add much value since it's really easy to build since we don't have many icons.

legopiraat:
Ready for review :)

packages/mutation-testing-elements/src/components/svg/index.ts
```diff
@@ -1,10 +1,47 @@
...
+
+  private readonly svgMapping: Map<string, SVGTemplateResult> = new Map();
+
+  constructor() {
+    this.svgMapping.set('cs', this.csharp);
+    this.svgMapping.set('html', this.html);
+    this.svgMapping.set('java', this.java);
+    this.svgMapping.set('js', this.js);
+    this.svgMapping.set('scala', this.scala);
+    this.svgMapping.set('typescript', this.typescript);
```
hugo-vrijswijk: Shouldn't this be `ts`? Unless I'm reading it wrong this is the extension

packages/mutation-testing-elements/src/components/svg/index.ts
```diff
@@ -1,10 +1,47 @@
...
+  }
+
+  public getIconForFile(fileName: string): SVGTemplateResult {
+    const fileExtention = this.getFileExtention(fileName);
```
hugo-vrijswijk: Should be 
```suggestion
    const fileExtension = this.getFileExtension(fileName);
```

packages/mutation-testing-elements/src/components/svg/index.ts
```diff
@@ -1,10 +1,47 @@
...
+    } else {
+      return this.file;
+    }
+  }
+
+  public getIconForFolder(): SVGTemplateResult {
+    return this.directory;
+  }
+
+  private getFileExtention(fileName: string) {
```
hugo-vrijswijk:
```suggestion
  private getFileExtension(fileName: string) {
```

packages/mutation-testing-elements/src/components/svg/index.ts
```diff
@@ -1,10 +1,47 @@
...
+    this.svgMapping.set('html', this.html);
+    this.svgMapping.set('java', this.java);
+    this.svgMapping.set('js', this.js);
+    this.svgMapping.set('scala', this.scala);
+    this.svgMapping.set('typescript', this.typescript);
+  }
+
+  public getIconForFile(fileName: string): SVGTemplateResult {
+    const fileExtention = this.getFileExtention(fileName);
+    const svg = this.svgMapping.get(fileExtention);
```
hugo-vrijswijk:
```suggestion
    const svg = this.svgMapping.get(fileExtension);
```

legopiraat: At least I was consistent 😅

packages/mutation-testing-elements/src/components/svg/index.ts
```diff
@@ -1,10 +1,47 @@
-import { svg } from 'lit-html';
+import { svg, SVGTemplateResult } from 'lit-html';
 
-// Octicons - github
-const file = svg`<svg aria-label="file" class="octicon octicon-file" viewBox="0 0 12 16" version="1.1" width="12" height="16" role="img"><path fill-rule="evenodd" d="M6 5H2V4h4v1zM2 8h7V7H2v1zm0 2h7V9H2v1zm0 2h7v-1H2v1zm10-7.5V14c0 .55-.45 1-1 1H1c-.55 0-1-.45-1-1V2c0-.55.45-1 1-1h7.5L12 4.5zM11 5L8 2H1v12h10V5z"></path></svg>`;
-const directory = svg`<svg aria-label="directory" class="octicon octicon-file-directory" viewBox="0 0 14 16" version="1.1" width="14" height="16" role="img"><path fill-rule="evenodd" d="M13 4H7V3c0-.66-.31-1-1-1H1c-.55 0-1 .45-1 1v10c0 .55.45 1 1 1h12c.55 0 1-.45 1-1V5c0-.55-.45-1-1-1zM6 4H1V3h5v1z"></path></svg>`;
+export class SvgService {
```
hugo-vrijswijk: Tests? 😁

packages/mutation-testing-elements/src/components/svg/index.ts
```diff
@@ -1,10 +1,47 @@
...
 
-export {
-  file,
-  directory
-};
+  // Octicons - Github
+  private readonly file: SVGTemplateResult = svg`<svg aria-label="file" class="octicon octicon-file" viewBox="0 0 12 16" version="1.1" width="12" height="16" role="img"><path fill-rule="evenodd" d="M6 5H2V4h4v1zM2 8h7V7H2v1zm0 2h7V9H2v1zm0 2h7v-1H2v1zm10-7.5V14c0 .55-.45 1-1 1H1c-.55 0-1-.45-1-1V2c0-.55.45-1 1-1h7.5L12 4.5zM11 5L8 2H1v12h10V5z"></path></svg>`;
+  private readonly directory: SVGTemplateResult = svg`<svg aria-label="directory" class="octicon octicon-file-directory" viewBox="0 0 14 16" version="1.1" width="14" height="16" role="img"><path fill-rule="evenodd" d="M13 4H7V3c0-.66-.31-1-1-1H1c-.55 0-1 .45-1 1v10c0 .55.45 1 1 1h12c.55 0 1-.45 1-1V5c0-.55-.45-1-1-1zM6 4H1V3h5v1z"></path></svg>`;
+
+  // VS code icons - Github
```
hugo-vrijswijk: Could you also add a link to where these come from, in case people want to add more?

packages/mutation-testing-elements/src/components/mutation-test-report-totals/index.ts
```diff
@@ -18,6 +18,8 @@ export class MutationTestReportTotalsComponent extends LitElement {
   @property()
   public currentPath: string[] = [];
 
+  private readonly svgService: SvgService = new SvgService();
```
nicojs: Please remove the explicit typing here. It's not needed.

hugo-vrijswijk:
Also I noticed that the used icons are not the 'official' ones used in VS Code. `vscode-icons/vscode-icons` is third-party, just with a similar name. The 'standard' ones included in VS Code are Seti UI icons, included here: https://github.com/Microsoft/vscode/tree/master/extensions/theme-seti. Downside to this is that it is added as a font, which we probably don't want due to not needing every icon. 

SVG's are available here: https://github.com/jesseweed/seti-ui/tree/master/icons. But they don't have a fill color (mapping is [here](https://github.com/microsoft/vscode/blob/master/extensions/theme-seti/icons/vs-seti-icon-theme.json), though). I personally think these are nicer than the vscode-icons, but they would need to have a fill.

packages/mutation-testing-elements/src/components/svg/index.ts
```diff
@@ -1,10 +1,47 @@
...
+    }
+  }
+
+  public getIconForFolder(): SVGTemplateResult {
+    return this.directory;
+  }
+
+  private getFileExtension(fileName: string) {
+    const tokens = fileName.split('.');
+    return tokens[tokens.length - 1];
```
nicojs: Why not `return fileName.split('.').pop()`. I thought you liked functional programming? ;)

packages/mutation-testing-elements/src/components/svg/index.ts
nicojs: Yep!

legopiraat:
@hugo-vrijswijk if I look at the icons I have in my own VScode these are the icons that are being used. So let's make a new issue if you don't like these icons after all.

packages/mutation-testing-elements/src/components/svg/index.ts
legopiraat: I DO like functional programming but I also like that my state does not get mutated like `pop()` does ;)

```ts
const data = [1, 2, 3, 4]

console.log(data) // [1, 2, 3, 4]
console.log(data.pop()) // 4
console.log(data) // [1, 2, 3]
```

nicojs:
image

Is the coloring correct? I find the contrast of JS and java a bit lacking...

nicojs:
Seems a bit better in VScode with white theme: 
image

hugo-vrijswijk:
@nicojs  I've updated the colour contrasts to be in line with VS Code:
image

PR #238, opened 2019-12-11
Add mutation testing ci/cd

nicojs:


packages/mutation-testing-metrics-scala/stryker4s.conf
```diff
@@ -0,0 +1,4 @@
+stryker4s {
```
hugo-vrijswijk: Stryker4s doesn't handle multi-module projects very well. See here for an example how to run it on multiple modules: https://github.com/stryker-mutator/stryker4s/pull/320/files#diff-c970abc4e2dc70e9336f54441c04c3dbR24-R26

PR #239, opened 2019-12-11
feat(ignored status): add MutantStatus Ignored

### LINKED ISSUE #85: Add state "ignored"
Both Stryker.NET as Stryker have (or will have) ways to ignore certain mutations (either by mutator name or other means). We want ignored mutations to also be visible in the mutation score. We should add this as a mutator state.

See https://github.com/stryker-mutator/stryker/issues/1472 

@Mobrockers what should the name be? I would prefer `Ignored` as it is really clear that the user choose to ignore it. Something like `skipped` could have any reason.

simondel:
Adds the MutantStatus `Ignored`. This state should be used whenever a user has chosen to ignore mutants.

These mutants are not counted in the score. They are added to the total number of generated mutants.

Fixes #85


hugo-vrijswijk:
Scala code LGTM

hugo-vrijswijk:
@simondel Can you also add a PR for the new status in the [Stryker handbook](https://github.com/stryker-mutator/stryker-handbook/blob/master/mutant-states-and-metrics.md) and add a example test json with an ignored mutant?

Mobrockers:
@hugo-vrijswijk @simondel @nicojs Seems like the docs were never updated with the new state, do we still want to update the docs?

PR #303, opened 2020-01-22
fix(metrics-scala): score should be NaN when there are 0 mutants

hugo-vrijswijk:
I might also update this PR to include it in JS metrics


richardwerkman:
I already added the NaN default to my PR for JS 😉

PR #313, opened 2020-01-29
Replace highlight.js with Prism

### LINKED ISSUE #30: Add line number support
As requested through stryker-net [#458](https://github.com/stryker-mutator/stryker-net/issues/458) we should add line numbers to the file view. This makes finding the mutation in the original source code easier.

hugo-vrijswijk:
Prism has a couple more features than just highlighting:

- 🔢 Line numbers! Closes #30
- 🌐 Autoload languages that aren't included by us (for example, if PHP is used, the highlighting style is retrieved from unpkg)!
- 📦 Small size! Even with these extra features, the build comes down to around the same size (or smaller if you remove autoload and line numbers).

The theme is slightly different, but we might even be able to find a theme that is very similar to the previous one, if wanted.

Before:
image

After:
image


packages/mutation-testing-elements/package.json
```diff
@@ -26,5 +26,9 @@
   "devDependencies": {
     "mutation-testing-metrics": "^1.2.2",
     "mutation-testing-report-schema": "^1.2.0"
+  },
+  "dependencies": {
+    "@types/prismjs": "^1.16.0",
```
nicojs: `@Types` are always dev-dependencies

packages/mutation-testing-elements/package.json
```diff
@@ -26,5 +26,9 @@
...
+    "prismjs": "^1.19.0"
```
nicojs: Is this really a dependency? We're not bundling it with webpack?

packages/mutation-testing-elements/src/components/mutation-test-report-file/index.scss
```diff
@@ -17,12 +17,6 @@
...
-code.hljs {
-    overflow: visible;
-    height: 100%;
-}
-
 pre {
-    overflow: visible;
-}
\ No newline at end of file
+    overflow-y: hidden !important;
```
nicojs: Do we realy need an `!important`? Also: why is overflow `hidden` (maybe add a comment to explain)?

packages/mutation-testing-elements/package.json
nicojs: EDIT: Since we're importing from `prism`, I guess this is bundled. Please make `prismjs` itself a dev dependency

packages/mutation-testing-elements/src/components/mutation-test-report-mutant/index.scss
```diff
@@ -4,3 +4,7 @@
 .disabled-code {
   text-decoration: line-through;
 }
+
+span.badge {
+    text-shadow: none;
```
nicojs: different formatting used. Apparently we're missing a formatter?

packages/mutation-testing-elements/src/typings/prism.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'prismjs/components/prism-core' {
+  // Core has same exports as root prismjs export
```
nicojs: Please also add this to a PR here: 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/prismjs/index.d.ts

hugo-vrijswijk:
> A few questions:
>    1. What was the primary reason to switch to Prism? I like the Line numbers, was that it?

Mostly line numbers, yes. It seems to have a lot more functionalities beyond just highlighting that we could use. I stumbled upon it when seeing a lot of the hip new websites use it (namely Vue & React documentation).

>    2. On the Prism website, I read: "Any pre-existing HTML in the code will be stripped off".
    How is it that our badges (that point out mutants) still work?

Yeah I had that same question for a while and initially it didn't work. Prism has a [keep-markup](https://prismjs.com/plugins/keep-markup/) plugin which needs to be used. I'll clarify that with some comments

packages/mutation-testing-elements/package.json
hugo-vrijswijk: I'll change both to devDependencies

packages/mutation-testing-elements/src/components/mutation-test-report-mutant/index.scss
hugo-vrijswijk: I can add Prettier in a different PR if you like? (Maybe after some of the other open PR's are merged 😃)

packages/mutation-testing-elements/package.json
hugo-vrijswijk: Added in 1007d2d7a09e05980e7b5232d3ef014075ae165d

packages/mutation-testing-elements/src/components/mutation-test-report-file/index.scss
hugo-vrijswijk: Changed it to `visible` now for the popup. 
image

> Do we realy need an !important?

Unfortunately, prismjs has a very specific selector of `pre[class*="language-"]` which takes precedence over just `pre`. So it's either a more specific selector like `pre.line-numbers` (which I don't like because it is so specific) or an important. What do you think? Maybe an id?

PR #326, opened 2020-02-03
chore(build): update npmignores

hugo-vrijswijk:
Too much was being published


packages/mutation-testing-metrics-scala/.npmignore
```diff
@@ -0,0 +1,6 @@
+**/*
```
nicojs: Why does this need a npmignore?

hugo-vrijswijk: It's still being published on npm unfortunately. I haven't found a way to fix it while still having lerna start the sbt publishing process. So at least now it doesn't publish all the jars to npm

PR #327, opened 2020-02-03
chore(build-elements): build module es2015 for elements

hugo-vrijswijk:


nicojs:
Why is this needed?

hugo-vrijswijk:
@nicojs Changing the elements tsconfig to use esmodules instead of commonjs allows webpack to remove boilerplate and optimize its tree shaking a lot more. This reduces the outputted file from 223KiB to 110KiB (halved!). 

Technically it's a breaking change because of the different module output. However, I think that's fine because browser-wise anyone who supports web-components will also support es-modules. For build systems it would be breaking, but I am not sure if anyone actually uses it that way

PR #330, opened 2020-02-04
chore: migrate to eslint & prettier

nicojs:
* 📏 Add typescript eslint plugin for type-aware linting
* 💄 Add prettier & prettier eslint plugin
* 🎨 Add prettier linting for json
* 👴 Remove tslint
* 🛠 Fix all lint issues


nicojs:
Let's merge this after your recent changes @hugo-vrijswijk

package.json
```diff
@@ -6,10 +6,11 @@
...
   "scripts": {
     "clean": "rimraf packages/*/{dist,dist-test}",
-    "all": "npm run lint && npm run build && npm run test",
+    "all": "npm run clean && npm run lint && npm run build && npm run test",
     "postinstall": "lerna bootstrap && link-parent-bin",
-    "lint": "tslint -p tsconfig.lint.json",
-    "prebuild": "npm run clean",
+    "lint": "npm run lint:json && npm run lint:ts",
+    "lint:ts": "eslint --ext .js,.ts --ignore-path .gitignore .",
+    "lint:json": "prettier --check .eslintrc *.json",
```
hugo-vrijswijk: Can we also lint scss files?

packages/mutation-testing-elements/package-lock.json
```diff
@@ -1,7 +1,30 @@
...
+			"version": "2.5.0",
+			"resolved": "https://registry.npmjs.org/@stryker-mutator/api/-/api-2.5.0.tgz",
+			"integrity": "sha512-6z2AZeGcI+ZzqZEBX6h+vWt11uBzLfAg24kV0b/CJZ69zNELRLkjJnLaE4aU1o5uE6OyFGKJJ0OVxtZrnvZSuQ==",
+			"dev": true,
+			"requires": {
+				"mutation-testing-report-schema": "^1.1.0",
+				"tslib": "~1.10.0"
+			}
+		},
+		"mutation-testing-report-schema": {
```
hugo-vrijswijk: Is this intended?

nicojs: I've added it in JS Doc here: https://github.com/stryker-mutator/mutation-testing-elements/blob/chore/eslint/packages/mutation-testing-elements/tsconfig-transpiler.js#L13

But we could remove it as well. Not really needed.

package.json
nicojs: Yes, let's add that in a separate PR. This one is already kind of big

nicojs:
Don't blindly merge this PR! One unit test is luckily failing (with good reason). The whitespace formatting within the `hmtl` template tags breaks the reporter:

image

packages/mutation-testing-elements/package-lock.json
hugo-vrijswijk: I meant the mutation-testing-report-schema dependency. I thought Lerna was supposed to take care of that instead of the published package being in the package-lock

nicojs:
Note: we need to wait until _at least_ https://github.com/prettier/prettier/pull/7208 is released.

We should test it out. Almost all our lit-html template code is whitespace significant. We don't want to break the html reporter.

hugo-vrijswijk:
Could we migrate to eslint first, and do prettier later?

nicojs:
Yes we could

nicojs:
Prettier 2 is released 🎉 
https://prettier.io/blog/2020/03/21/2.0.0.html

I've updated and everything looks splendid! Fill merge when the build is green

PR #518, opened 2020-06-07
Improve performance of report-mutant rendering

### LINKED ISSUE #423: Html report stuck when the amount of data is too large
Hi, there. Thanks for your excellent work, my team is going well with stryker recently, and we are trying to recommand it to other teams.
We've also met some trouble with the html report feature recently. That is: when the json data is too large, the whole report page would stuck. 
For example, a js file which contains nearly 1300 mutants would stuck for about 15-20s or even more. Another file with 4800+ mutants(I know this file is unreasonably large.....) got stucked for too long that chrome suggest to close it because of no responding.
Now I try to display  the ' on-loading' tag when some big file is loading, but if too large file(like I mentioned before) is clicked, the whole page would stuck and I have to close it.
I wonder if it is possible to add some feature such as lazy loading to solve this problem?
Thanks for you time any way.

hugo-vrijswijk:
(Hopefully) partially fixes #423

I noticed that if there are multiple mutations on the same location the `mutation-test-report-mutant` was rendered in a nested structure. So each report-mutant element is rendered **inside** the one that went before it. This makes the whole report nested, and with a large amount of mutants on the same location this has a big impact on performance.

With this change report-mutants are rendered in a flat structure, instead of nested for each mutant. The browser is much more efficient at rendering this. And it looks like lit-html deals with it a bit more efficiently, too.

I also cleaned up the rendering code a little to no longer add or close spans when it shouldn't. So we don't have to rely on the mercy of how browsers decide to clean up our invalid HTML.

### Previous:

image

Etcetera. You get the picture.

### New:

image

In Chromium Edge, this speeds up a report with 1000 mutants on the same location from ~50 seconds of an unresponsive white page to ~9.

EDIT: Unfortunately this breaks the line-through when you open a mutant. I'll see if I can fix that

### DIFF
--- a/packages/mutation-testing-elements/src/lib/codeHelpers.ts
+++ b/packages/mutation-testing-elements/src/lib/codeHelpers.ts
@@ -26,19 +26,16 @@ export function renderCode(model: FileResult): string {
     startedMutants.push(...currentMutants);
 
     const builder: string[] = [];
-    if (currentMutants.length || mutantsEnding.length) {
-      currentMutants.forEach(backgroundState.markMutantStart);
-      mutantsEnding.forEach(backgroundState.markMutantEnd);
-
-      // End previous color span
+    if (mutantsEnding.length) {
       builder.push('</span>');
+      mutantsEnding.forEach(backgroundState.markMutantEnd);
+    }
 
-      // End mutants
-      mutantsEnding.forEach(() => builder.push('</mutation-test-report-mutant>'));
+    if (currentMutants.length) {
+      currentMutants.forEach(backgroundState.markMutantStart);
 
-      // Start mutants
-      currentMutants.forEach((mutant) => builder.push(`<mutation-test-report-mutant mutant-id="${mutant.id}">`));
-
+      // Write mutants
+      currentMutants.forEach((mutant) => builder.push(`<mutation-test-report-mutant mutant-id="${mutant.id}"></mutation-test-report-mutant>`));
       // Start new color span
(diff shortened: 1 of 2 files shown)


nicojs:
Thanks for looking into this @hugo-vrijswijk. 

I've implemented the nesting of `<mutation-rest-report-mutant>` elements by (my own) design. That way, the correct code is crossed out. See this C# example:

**With nesting (master branch)**
image

**Without nesting (improve-rendering-performance branch)**
image

🤷‍♂️

hugo-vrijswijk:
Yeah that was the only issue I have with my changes. I think it's worth looking into having the text crossed-out without the nesting

nicojs:
I think we should first look at other low hanging fruit. For example, right now each mutant component is drawn on the screen twice. Once before prism highlights the code and once after. This is done because prism (or highlightjs) totally rewrites the HTML. After that is done, the correct mutant is rebound here: https://github.com/stryker-mutator/mutation-testing-elements/blob/4c835a8e456cfc4502388744c22b3276e8eef197/packages/mutation-testing-elements/src/components/mutation-test-report-file/index.ts#L123

See also: https://github.com/PrismJS/prism/issues/2275

Mobrockers:
@nicojs @hugo-vrijswijk Do we actually want the cross-out text? In most cases it makes it harder for me to detect what the original code was. Perhaps annotating with red is enough clarification?

hugo-vrijswijk:
Closing for now. Might take another look at this some time in the future

PR #703, opened 2020-10-26
test(screenshots): add screenshot test mechanism

nicojs:
Add tests to verify that the report does not change during normal ci/cd updates. 

* Run them from the `mutation-testing-elements` package (you can also use vscode's launch.json configuration).
* `npm run test:integration` to run them.
* `npm run test:integration:update` to update them.
* When the diff comparison fails, it will output a `.diff.png` file pointing out the differences: 
    
    
* When the diff comparison fails in the workflow, the diffs are uploaded as artifacts.
    image

Couple of points we need to figure out:
1. They are OS-dependent. Both Windows and Linux screenshots are checked in and compared in the pipeline.
1. They are headless vs non-headless dependent. When running without a headless browser, the tests are skipped, because it would always fail.

An example of not honoring one of these 2 rules:
image

You can update the screenshot for both Linux and Windows by running the "Update screenshots" workflow on your branch (with github). _Use with caution_ always double-check the diff before updating!


nicojs:
@hugo-vrijswijk  👀

nicojs:
> Couple of points we need to figure out:
> 
> 1. They will fail on another os than Linux (or WSL2, maybe mac)
> 1. They will fail on other browsers than chrome headless.

I've currently decided to simply skip the tests on windows. If this is an issue going forward we should improve it. For now, this is already better than we currently have at least.

packages/mutation-testing-elements/test/integration/setup/setup-chai.ts
```diff
@@ -0,0 +1,53 @@
...
+import compareImages from 'resemblejs/compareImages';
+import type { ResembleComparisonResult } from 'resemblejs';
+import { expect } from 'chai';
+
+let currentSnapshotFile: string;
+const reportsDir = path.resolve('reports', 'diff');
+
+chai.use(({ Assertion }) => {
+  Assertion.addMethod('matchScreenshot', async function () {
+    const snapshotFileExists = existsSync(currentSnapshotFile);
```
hugo-vrijswijk, line 17: This function is async, so why the sync exists?

PR #805, opened 2020-12-18
feat(schema): add config, test coverage and metadata

### LINKED ISSUE #36: Show testrun result for each mutant
It would be nice if we could show why a mutant was marked as killed by displaying the testrun output. That way users can see what test case killed the mutant.

This could also help by detecting false positives. When the mutation score is much higher than expected it could be seen that many mutants have the same testrun output (for example a bad unit test that fails randomly)

We could show the testrun output when clicking on the mutant.

### LINKED ISSUE #720: Allow for test framework and test framework version keys in schema
In stryker.net we have a feature we call dashboard compare. You can use it to pull a past mutation report from a storage location and use that result to fill in any mutants in code that has not been changed in order to speed up test runs.

I would like to be able to reject a past mutation report if it was generated using a different version of stryker.NET (to prevent incorrect results due to changes between stryker versions), as well as reject the report if it was not created by stryker.net (to prevent unneccesary processing when we know there is no way for the report to be valid for stryker.net).

@nicojs @hugo-vrijswijk do you guys see any issues adding this information to the report?

### LINKED ISSUE #786: Add performance statistics, tests, system and config to schema
This issue combines other issues to create a coherent design of changes we want to make to the schema.

#720, #36 , https://github.com/stryker-mutator/stryker/issues/2647

We want to add the following information to the schema:
1. Mutation test framework configuration used for the run.
1. Tests in the test suite and how they relate to mutants (mutant coverage and "killed-by").
1. Performance statistics for the phases (setup, initial run, mutation testing), as well as performance per mutant.
1. System information of the machine that generated the report.

**Note**: We want to keep the schema _free of derived values_. Only the actual data. Derived values will still be available through the mutation-testing-metrics dependency (delivered as jar and npm package).

This example will be kept up to date with the discussion below.

```js
{
  "config": { // object - optional
    "testRunner": "jasmine" // free object format allowed here
  },
  "testFiles": { // optional
    "test/foo.spec.js": { // If the mutation testing framework cannot determine the file name, it can use an empty string as key here
        "tests": [{
          "id": "test-1"  // test ID
          "name": "foo should be bar", // required
          "location": { // location in the file, optional
            "start": {        // start position in the file, required
              "line": 4,
              "column": 6
            },
            "end": {        // end position of the test, optional
              "line": 12,
              "column": 1
            }
        }]
      }
    }
  },
  "files": {
    "test.js": {
      "mutants": [ // left out current fields for brevity 
        {
          "id": "1", 
          "static": false, // optional, marks the mutant as static/non-static
          "coveredBy": ["test-1", "test-2"], // Optional
          "killedBy": [ "test-1"], // Optional
          "testsRan": 1, // Optional, can defer from testCoverage because of "--bail" mode in test runners.
          "duration": 42 // in ms - optional
        }
      ]
    }
  },
  "performance": { // optional
    "setup": 31000, // in ms, required, fill with `0` if this run didn't contain a "setup" step
    "initialRun": 31000, // in ms, required, fill with `0` if this run didn't contain a "initialRun" step
    "mutation": 1233500 // in ms, required, fill with `0` this run didn't contain a "mutation" step
  },
  "system": {  // optional
    "ci": false, //  Required. `true` if the run was performed on a CI like Github Actions.
    "os": { // optional
      "version": "Windows 10 Pro", // optional
      "platform": "win32", // optional
      "release": "10.0.19041" // optional
    },

    "cpu": { // optional
      "model": "Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz", // optional
      "logicalCores": 16, // required
      "baseClock": 2304 // optional
    },

    // RAM can have an impact on performance when there isn't enough.
    "ram": { // optional
      "total": 34.07 // required, in MB
    }
  },
  "framework": {
    "name": "Stryker", // required
    "version": "4.2.0", // optional
    "branding": { // optional
      "homepageUrl": "https://stryker-mutator.io", // required
      "imageUrl": "https://stryker-mutator.io/images/stryker.svg" // optional, can be a data-url
    },
    "dependencies": { // optional
      "@stryker-mutator/core": "4.2.0", // key-value pair.
      "@stryker-mutator/jest-runner": "4.2.0",
      "jest": "26.6.1" ,
      "node": "14.15.1"
    }
  }
}
```

nicojs:
Fixes #786 
Fixes #720
Fixes #36


packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
```diff
@@ -6,12 +6,42 @@
...
+        "type": "object",
+        "title": "TestDefinition",
+        "description": "A test in your test suite that ran during mutation testing.",
+        "required": ["name"],
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "Name of the test, used to display the test."
+          },
+          "fileName": {
```
Mobrockers, line 32: Wouldn't that be filePath if it's relative to the project root?

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
```diff
@@ -73,6 +103,18 @@
                   "description": "Unique id, can be used to correlate this mutant with other reports.",
                   "examples": ["321321"]
                 },
+                "description": {
+                  "type": "string",
+                  "description": "Description of the applied mutation.",
+                  "examples": ["removed call to java/io/Writer::write"]
+                },
+                "killedBy": {
```
Mobrockers, line 111: What if we cannot determine the exact test, or if there are multiple that kill this mutant?

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
```diff
@@ -83,24 +125,140 @@
...
+          "description": "Time it took to run the mutation test phase in milliseconds."
+        }
+      }
+    },
+    "system": {
+      "type": "object",
+      "title": "SystemInformation",
+      "description": "Information about the system that performed mutation testing",
+      "properties": {
+        "ci": {
```
Mobrockers, line 178: We should be careful with this value imo. The only possible values would be true or maybe-not. We can't definitively say that something did not run in a ci pipeline as we cannot support detection for all ci pipeline tools.

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
nicojs: It's optional, so if you cannot determine you simply leave it out. If there are multiple tests, I would suggest using the first one here? I think it is a good practice to use "bail" mode for mutation testing, so stop testing as soon as one test fails.

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
nicojs: And "yes" is also unknown, since I can run with `CI=TRUE` in my environment variables. I personally don't mind myself, I think it stands to reason that this value is "best-effort". I can clarify that in the description if you want.

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
nicojs: Why would a `filePath` make it relative? I personally like fileName better. If frameworks choose to use the full name here I wouldn't mind. We'll probably just present it in the front report.

Mobrockers: The description specifies that it should be relative to the project root, that makes it a path and not a filename imo.

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
Mobrockers: Why not just list all tests that killed a mutant? Bail mode cannot be used in all mutation testing modes for stryker.net currently.

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
Mobrockers: I think it is a good idea to clarify that in the description so anyone using the schema can share the same definitions.

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
nicojs: Good point, changing it.

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
nicojs: It's mostly because I want to add a view later listing all tests that didn't kill a mutant. I want that list to not have false-negatives. So in my opinion every test should be able to kill unique mutants if that makes sense. If we would allow a list here, this new view might work differently depending on the way "killed-by" was reported.

I'm not strongly against it, just one man's opinion.

nicojs: We had a conversation about it and let's make it an array.

It's up to the mutation testing framework to decide to use "bail" mode or not, but that should still be a best practice IMO.

We've also discussed merging this array into the `testCoverage` array but decided to keep them separate. A mutation testing framework might not implement test coverage, but still want to report the `killedBy`.

packages/mutation-testing-report-schema/src/mutation-testing-report-schema.json
```diff
@@ -83,24 +125,140 @@
...
-                "location": {
-                  "$ref": "#/definitions/location"
-                },
                 "status": {
                   "type": "string",
+                  "title": "MutantStatus",
                   "description": "Result of the mutation.",
                   "enum": ["Killed", "Survived", "NoCoverage", "CompileError", "RuntimeError", "Timeout", "Ignored"]
+                },
+                "testCoverage": {
```
nicojs, line 134: Rename to `coveredBy` to be more in line with `killedBy`.

PR #871, opened 2021-02-02
feat(metrics): add test metrics

### LINKED ISSUE #864: Show information about killed by and covered by tests
I'm experimenting with ways to display the new test coverage and killed by information we've added in the schema with #805.

As you can see, this is not easy 😢

image

Our popover is simply not suited for this information. Of course our modal would be big enough, but I think this is horrendous for the user experience. There is a lot of value in information that is easy and natural to access.

 So we need a more out-of-the box approach. I was thinking of a "push drawer without overlay". Example: https://mdbootstrap.github.io/bootstrap-material-design/docs/4.0/material-design/drawers/#push

The ideau would be that the drawer would 'push in' from the right whenever you click a mutant. We could maybe replace _both the popover as the modal dialog_ with this new feature.

@hugo-vrijswijk @richardwerkman @Mobrockers @simondel and others, thoughts?

### LINKED ISSUE #872: Naming the metrics for tests
While we're implementing the details about tests I want to make sure we get the terms for those metrics just right. I also want to know exactly what we want to "count" and show to the end-user in the overview table. Here is my proposal:

(all of these are per file/directory)

* `total`:  The total number of tests.
* `pacifist`:  The total number of tests that didn't kill any mutants.
* `without coverage`: total number of tests that didn't even cover a single mutant.

@hugo-vrijswijk @Mobrockers what do you think? Am I forgetting important metrics here? What about the names?

nicojs:
Add metrics about killed by and covered by tests to mutation-testing-elements.

drawer-ftw

UI changes:

* Replace both the _modal dialog_ and the _popover_ with the new _drawer_ (from the bottom).
* The drawer can be _closed_, _half_ or _open_. 
  * When closed, it is not visible
  * When half, it only shows a summary. 
  * When open, it also shows details.
* The summary has a header and content. 
  * The header is the id, mutator name, status, and start location 
  * The content had 🎯 1 killed test, ☂ the number of covered tests, and the 📖 description (only if exists for all 3). 
  * The "Read more" button in the header is only visible when there is test information.
* The details contain a list of tests 
  * With "🎯" in front for killed.
  * With "☂" in front for covered.
  * Tests have a "title" attribute explaining about killed/covered (when hovering over with your mouse).
  * A scroll bar is shown when the number of tests gets too large.
* Integration tests updated, including screenshot snapshot tests in dark and light mode.

Closes #864

Metrics changes:

Implement new `calculateMutationTestMetrics`. It results in an object containing both the `systemUnderTestMetrics` and the `testMetrics`. 
* The test metrics contain of: `Not covering`, `Not killing`, `Killing` and `Total`. These are not yet shown.
* The resulting `mutant` entities possibly have `killedByTests`, `coveredByTests` pointing to the actual tests.
* The resulting `test` entities possibly have `killedMutants`, `coveredMutants` pointing to the actual mutants. (many to many)

Closes #872


nicojs:
@Mobrockers @hugo-vrijswijk 😇

Mobrockers:
Looks cool! I assume the description looks like that because Stryker used to put the killed by test in the description?

nicojs:
> Looks cool! I assume the description looks like that because Stryker used to put the killed by test in the description?

Exactly! We didn't have a place to put that info before, so I decided to put it in the description for now. 

fit

The description should from now actually contain the description of the mutation itself.

packages/mutation-testing-elements/src/style/prismjs.scss
```diff
@@ -1,4 +1,5 @@
 
 
 @import "./prismjs.css";
-@import "~prismjs/plugins/line-numbers/prism-line-numbers.css";
+// @import "~prismjs/plugins/line-numbers/prism-line-numbers.css";
```
hugo-vrijswijk, line 4: Why is this commented out?

packages/mutation-testing-metrics/src/model/test-model.ts
```diff
@@ -0,0 +1,32 @@
...
+  public addKilled(mutant: MutantModel) {
+    if (!this.killedMutants) {
+      this.killedMutants = [];
+    }
+    this.killedMutants.push(mutant);
+  }
+
+  constructor(input: TestDefinition) {
+    Object.entries(input).forEach(([key, value]) => {
+      // @ts-expect-error dynamic assignment so we won't forget to add new properties
```
hugo-vrijswijk, line 28: Won't we still have to any new properties to this class if they are added? I'm unsure what extra work this is preventing

packages/mutation-testing-elements/src/components/mutation-test-report-app/theme.scss
```diff
@@ -16,7 +22,7 @@
   --bs-modal-content-border-color: #{$modal-content-border-color};
   --bs-modal-header-border-color: #{$modal-header-border-color};
```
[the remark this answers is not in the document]
nicojs: Yes. It's BS 👍

packages/mutation-testing-elements/src/style/prismjs.scss
nicojs: Good catch! I was playing around with prism diff's (which are incompatible with the line numbers plugin). 😅

packages/mutation-testing-metrics/src/model/test-model.ts
nicojs: I'm using this pattern 4 times in total. Each time it saves me from copying over some properties and making accidental copy-and-paste errors. 🤷‍♂️

nicojs:
image

It's working!!! It's working! [insert meme here] 💨⛑

packages/mutation-testing-metrics/src/model/test-model.ts
hugo-vrijswijk: Fair enough

hugo-vrijswijk (approved):

PR #893, opened 2021-02-11
feat(schema): add test file source and status reason

### LINKED ISSUE #891: Add a "statusReason" field
We currently don't have a place to put the error message for an error mutant, or the failure message for killed mutants. That is a shame because it might help the user. In Stryker, we currently put this in the description, which I think is not nice at all.

fit

Proposal:

add an optional `statusReason` string field. It _should_ have:
* The error message if the mutant status is error
* The failure message if the mutant is killed.

@Mobrockers @hugo-vrijswijk thoughts?

### LINKED ISSUE #892: Add a "source" field to test files
I want to add an optional `"source"` field to the test files. Comparable to the current `source` field to source files.

Filling this is pretty easy in Stryker. We can also use that for `--diff` mode, as well as use it in the report to show more details in the tests view.

@Mobrockers @hugo-vrijswijk thoughts?

nicojs:
Add `statusReason` to a mutant result and `source` to a test file.

Status reason:
The reason that this mutant has this status. In the case of a killed mutant, this should be filled with the failure message(s) of the failing tests. In case of an error mutant, this should be filled with the error message.

Closes #891 

Source:
Full source code of the test file. This can be used to display in the report.

Closes #892

### DIFF
--- a/packages/mutation-testing-metrics-scala/circe/src/main/scala/mutationtesting/circe.scala
+++ b/packages/mutation-testing-metrics-scala/circe/src/main/scala/mutationtesting/circe.scala
@@ -38,12 +38,13 @@ object circe {
     Codec.forProduct3("source", "mutants", "language")(FileResult.apply)(f => (f.source, f.mutants, f.language))
 
   implicit lazy val mutantResultCodec: Codec[MutantResult] =
-    Codec.forProduct10(
+    Codec.forProduct11(
       "id",
       "mutatorName",
       "replacement",
       "location",
       "status",
+      "statusReason",
       "description",
       "coveredBy",
       "killedBy",
@@ -56,6 +57,7 @@ object circe {
         m.replacement,
         m.location,
         m.status,
+        m.statusReason,
         m.description,
         m.coveredBy,
         m.killedBy,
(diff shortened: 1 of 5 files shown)


nicojs:
@Mobrockers @hugo-vrijswijk 👀

hugo-vrijswijk (approved):
LGTM.

Small question because it still is not clear to me. When is something a description and not a statusReason? Do we think it should be a description of the mutation?

nicojs:
Yes, I think the `description` of a mutant should be describing the mutant itself. Pitest is a good example here. Instead of the `replacement`, it fills the `description`:

```js
 {
  {
    id: 18,
    mutatorName: 'VoidMethodCallMutator',
    description: 'removed call to org/pitest/elements/models/MutationTestSummaryData::addTestSummary',
    location: { start: { line: 23, column: 1 }, end: { line: 23, column: 36 } },
    status: 'Survived'
  },
  {
    id: 19,
    mutatorName: 'ReturnValsMutator',
    description:
      'mutated return of Object value for org/pitest/elements/models/PackageSummaryData::getSummaryData to ( if (x != null) null else throw new RuntimeException )',
    location: { start: { line: 28, column: 1 }, end: { line: 28, column: 64 } },
    status: 'Killed'
  },
}
```

PR #911, opened 2021-02-21
feat(metrics-scala): add config to MutationTestResult

hugo-vrijswijk:


nicojs (approved):
Looks great! Is it possible to assign a default to a type parameter? Otherwise a default of `Nothong` might make sense (for when you don't care about config)

hugo-vrijswijk:
> Is it possible to assign a default to a type parameter? Otherwise a default of Nothong might make sense (for when you don't care about config)

Can't do default type parameters in Scala. You can still set it to `Any` if you don't care about the value inside. I think this makes sense because there can still be a value inside, you just don't care about what type it is

PR #920, opened 2021-02-23
fix(schema): regression: allow import of schema

nicojs:


nicojs:
No longer needed. I was experiencing some strange behavior this morning, schema json file wasn't copied to the output dir. Now it is again 🤷‍♂️

PR #947, opened 2021-03-05
feat(dark-mode): update dark mode colours

hugo-vrijswijk:
Updates the dark mode colours and add a new code highlighting theme. The colours are based on the current Stryker website. The code theme is similar to Base16-eighties. 

image

image

image

Some other changes:

- The score progress bar text colour is now calculated based on whatever contrast works best (black for orange, for example)
- Light code theme is also slightly updated (see snapshot image)
- Background colour of a mutant (the green highlighted code after a killed mutant, for example) is now opacity-based instead of a hardcoded colour. Works much nicer IMO
- Drawer has a slightly different colour to the background colour to make it more distinct
- Clean up some CSS classes using generated SCSS loops
- Give the table slightly more air to breathe (up for debate?)
- Vertically-align the navbar items


packages/mutation-testing-elements/testResources/simple-example/index.html
```diff
@@ -11,11 +11,11 @@
   <script>
     const app = document.getElementsByTagName('mutation-test-report-app').item(0)
```
nicojs, line 12: Maybe it makes sense to add the theme color as a parameter of the event? For example: `{ themeColor: "#18191a" }`? Otherwise, the application needs to guess what the color is.

hugo-vrijswijk: Yes I thought about that too and it's not too difficult, but the problem is when you first open the report no event is thrown to catch the initial background color

nicojs: Ah yes. Then it makes sense to also reflect this on the theme property here:

https://github.com/stryker-mutator/mutation-testing-elements/blob/d117561a83bf93a8d884d5fd6d84aa1c7918a4ae/packages/mutation-testing-elements/src/components/mutation-test-report-app/mutation-test-report-app.component.ts#L36-L37

That way you can either look at the property or look at the event data.

nicojs:
Love the way this looks btw 😍

packages/mutation-testing-elements/testResources/simple-example/index.html
hugo-vrijswijk: Added `themeBackgroundColor` as a property

packages/mutation-testing-elements/src/components/mutation-test-report-app/mutation-test-report-app.component.ts
```diff
@@ -39,6 +39,11 @@ export class MutationTestReportAppComponent extends LitElement {
   @property()
   public drawerMode: DrawerMode = 'closed';
 
+  @property({ attribute: false })
+  public get themeBackgroundColor(): string {
+    return getComputedStyle(this).getPropertyValue('--bs-body-bg');
```
nicojs, line 44: Cool, didn't know you could read a property value.

packages/mutation-testing-elements/src/components/mutation-test-report-app/theme.scss
```diff
@@ -58,35 +59,91 @@
...
+  --prism-attr: #e3116c;
+  --prism-string: var(--prism-attr);
+
+  --prism-keyword: #00009f;
+  --prism-selector: var(--prism-keyword);
+  --prism-tag: var(--prism-keyword);
 }
 :host([theme='dark']) {
-  --prism-maintext: #ccc;
+  // From https://atelierbram.github.io/syntax-highlighting/prism/demo/assets/css/prism-base16-eighties.dark.css
```
nicojs, line 108: So in the end you didn't use github dark theme?

packages/mutation-testing-elements/src/style/bootstrap/_darkly-variables.scss
```diff
@@ -5,186 +5,187 @@
 // Color system
 //
 
-$darkly-white:    #fff !default;
```
nicojs, line 8: I don't mind the new formatting, but maybe we need to introduce prettier as a lint task to make sure we keep this consistent

packages/mutation-testing-elements/README.md
```diff
@@ -106,6 +119,10 @@ Default: _user preference (OS setting)_
 
 Specify in which theme the report needs to be shown. Possibilities: `'light'` or `'dark'`.
 
+### `themeBackgroundColor` [`string`]
+
+Read-only property with the hex code of the background-color from the current theme. 
+
 ### ⚡ `theme-changed` [`CustomEvent<{ theme: string }>`]
```
nicojs, line 126: Maybe we can add `{ themeBackgroundColor: string }` here as well for consistency?

packages/mutation-testing-elements/src/style/bootstrap/_darkly-variables.scss
hugo-vrijswijk: Yeah I can add that in a separate PR (after) to keep the diff for this PR smaller

packages/mutation-testing-elements/src/components/mutation-test-report-app/theme.scss
hugo-vrijswijk: No, this is closer to [this theme](https://github.com/chriskempson/tomorrow-theme) which is what the original dark theme was based on, but discontinued

packages/mutation-testing-elements/README.md
hugo-vrijswijk: Good idea! Added it

packages/mutation-testing-elements/src/components/mutation-test-report-app/theme.scss
```diff
@@ -58,35 +59,91 @@
...
-  --prism-strings: #690;
-  --prism-namespace: slategray;
-  --prism-names: #905;
-  --prism-symbol: #dd4a68;
-  --prism-url: #67cdcc;
+  --prism-maintext: #393a34;
+  --prism-background: #f6f8fa;
+  --prism-border: #ddd;
+
+  // from https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/themes/github.js
```
nicojs, line 72: Is this still correct?

PR #996, opened 2021-03-26
docs(states and metrics): document test metrics

nicojs:
Documents what has been decided in #872


hugo-vrijswijk:
This is already implemented in metrics, right?

nicojs:
Yes, but it's missing on the website currently.

hugo-vrijswijk (approved):

PR #1014, opened 2021-04-06
feat(test-view): add test view with test details

### LINKED ISSUE #865: Show information about tests
I want to create a "test file view" in the report. My current idea is to show the test files in the normal 

Some inspiration:

image
image
image

If we go with a combined files table like this, then we will need to find a solution for the metrics inside the table. You might think we could simply show 2 tables back to back, 1 for the files under test and 1 for the test files. However, it is very common in JavaScript to put your unit test files _right next to your source files_, so simply splitting them out into 2 tables might not work, since then you might see 1 directory twice (once in the table with metrics for tests and once in the table for metrics for files under test). I'd personally prefer a combined view if we can manage it.

Some more thoughts:

1. If of course, the test file paths are unknown, we should add a "tests" (or something) at the bottom of the table with a view for all tests.
2. I would prefer to end up with the _actual code of the test files_ and inlining information there (as we do for mutants). I think that is the most user-friendly in the end. This would be possible for most JavaScript test runners.

nicojs:
Add a "🧪 Test" view to the mutation test report. In this view, you can find information about tests.

<table>
<tbody>
<tr><td>

</td>
<td></td>
<td></td></tr>
<tr>
<td></td>
<td>
</td>
<td>

</td>
</tr>
</tbody>
</table>

It functions the same way as the already existing view, which is now named the "👽 Mutant" view.

When there is information about tests in the report JSON, `mutation-testing-elements` will now render 2 nav tabs (top left). Here you can switch between the 2 views. The mutant view will open by default.

image

When opening the test view, you either get the file/directory view or a big list of all tests, depending on the data available. 

* If the report JSON contains tests without a test file (i.e. `{ "testFiles": { "": { "tests": [/*...*/] } }`), it will show a big list of tests
* If the report JSON contains the test file names (i.e. `{ "testFiles": { "foo.spec.js": { "tests": [/*...*/] } }`), it will show the directory view

When navigating to a file view, either the source of that file is displayed with the tests in them, or a list of tests is displayed when there is no source available.

Closes #865


packages/mutation-testing-elements/src/components/mutation-test-report-app/mutation-test-report-app.component.ts
```diff
@@ -74,6 +85,11 @@ export class MutationTestReportAppComponent extends LitElement {
         this.theme = 'light';
       }
     }
+
+    // Set the default view to "mutant" when no route is selected
+    if (this.path.length === 0 || (this.path[0] !== View.mutant && this.path[0] !== View.test)) {
+      window.location.hash = `#${View.mutant}`;
```
hugo-vrijswijk, line 91: I noticed that this replaces the browser history, which means you need to press back twice to go to the previous page whenever the page loads. A better alternative could be:

```suggestion
      window.location.replace(toAbsoluteUrl(`${View.mutant}`));
```

packages/mutation-testing-elements/src/style/globals.scss
```diff
@@ -0,0 +1,3 @@
+// Global variables loaded in mutation-test-report-app
+
+$mte-drawer-height-half-open: 120px; // Leave room for 2 lines of text
```
hugo-vrijswijk, line 3: Shouldn't this be a CSS variable in theme.css?

hugo-vrijswijk (commented):
@nicojs I looked through most of it, though I skipped over the tests (I'm sure they're fine), just 1 small remark. Otherwise, looks great! Feel free to look at my comments and last commit and merge if you agree 👍

packages/mutation-testing-elements/src/style/globals.scss
nicojs: No, it is a hardcode variable used in a couple of sass files. I didn't want to pollute CSS variables where we're not switching the values per theme. I thought to a group such settings in `globals.scss`. I'm open to suggestions 🤷‍♀️.

PR #1065, opened 2021-04-29
chore(rename): rename NotKilling -> Covering

### LINKED ISSUE #1041: New name for the "NotKilling" status
I've implemented the test view in #1014 , however I don't think the view is very clear with the "NotKilling" column there.

(Note, feedback on UI etc plz in the PR and not here)

I would personally expect that `Killing` + `NotKilling` would result in `Total tests`. However, that's not the case because of `NotCovering` not being included in `NotKilling`. I think we should really find a better word for `NotKilling`.

I personally like `Pacifist` better. It sounds different enough from `Killing` and `NotCovering` to understand what's going on. I know that the word "Pacifist" isn't all that common, but I do think it is growing in popularity with the rise in popularity of "nerd culture". 

I'm open to other suggestions as well, but everytime I think about it I fall back to pacifist 🤷‍♂️

Note: no matter what we choose, we should include popover or useful links that explain what everything means.

nicojs:
Fixes #1041


hugo-vrijswijk (approved):

hugo-vrijswijk:
My beautiful peace dove 😿

nicojs:
Well... a dove 🕊 doesn't make a lot of sense when you call the status "Covering", right?

packages/mutation-testing-elements/src/lib/htmlHelpers.ts
```diff
@@ -58,8 +58,8 @@ export function getEmojiForTestStatus(status: TestStatus) {
   switch (status) {
     case TestStatus.Killing:
       return '✅';
-    case TestStatus.NotKilling:
-      return '🕊';
+    case TestStatus.Covering:
+      return '☂';
```
PauloGoncalvesBH, line 62: great choice of emoji, I liked it haha

hugo-vrijswijk:
> Well... a dove 🕊 doesn't make a lot of sense when you call the status "Covering", right?

Clearly the time for peace is over 🪓

PR #1069, opened 2021-04-30
feat(info): add tooltip information about states

### LINKED ISSUE #1060: Add a tooltip with information button to explain mutant states
We should add a tooltip with an information button to explain all mutant states, test states, and mutant settings (like `static`, etc).

Simple mockup for illustration:

image

The information button (not in the mockup) could link to: https://stryker-mutator.io/docs/mutation-testing-elements/mutant-states-and-metrics

nicojs:
Add "title" attributes to all mutant states and test states, explaining them in more detail. Also, add a link to the "mutation states and metrics page".

image

image

Refined the "mutation states and metrics" page to be a bit more clear.

Closes #1060


nicojs:
@PauloGoncalvesBH 
@hugo-vrijswijk 
@rouke-broersma 
👀 

This is a simple, but I hope effective implementation. Since I'm using the 'normal' [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) attribute, the browser natively does the positioning for us. It works pretty well. 

I've aligned the information in the tooltips with our "mutation states and metrics" page.

I've also added a ℹ-icon with a link to the page itself.

👍/👎?

rouke-broersma:
Not sure on the position of the [i] but title is good so 👍👍👍

nicojs:
Ok, I've merged this. I'd be happy to change the location of the ℹ-icon if a better suggestion comes in. 😄

PR #1073, opened 2021-05-03
fix(elements): add check if localStorage is available

### LINKED ISSUE #1064: Fallback for localStorage unavailable
image

Access to localStorage can be denied by the environment where `mutation-testing-elements` is hosted. This kind of makes sense, I just never realized this would be the case.

We should silently ignore these kind of errors, storing the preferred theme should be a best-effort feature.

hugo-vrijswijk:
Fixes #1064

Checks if localStorage exists. If it exists but it is disabled, the browser will throw an error. So we can't just check if `localStorage !== undefined`


nicojs:
Thanks! Looks great 👍

PR #1125, opened 2021-05-22
fix(elements): don't import report-schema in elements

hugo-vrijswijk:


packages/elements/src/components/theme-switch/theme-switch.scss
```diff
@@ -13,7 +15,7 @@ $width: 2.9em;
     }
   } @else if $exp < 0 {
     @for $i from 1 through -$exp {
-      $value: $value / $number;
+      $value: math.div($value, $number);
```
nicojs, line 18: Why is this included here?

packages/elements/src/style/bootstrap-reboot.scss
```diff
@@ -1,4 +1,6 @@
-@import 'bootstrap/scss/bootstrap-reboot.scss';
```
nicojs, line 1: Why are these css changes included?

nicojs (approved):
Wow great catch! Too bad this is a breaking change for typescript projects. I've tried to play around with webpack and ts configuration to make it so "schema" is tree shakeable, but without any luck, so this seems to be the only solution indeed. 

Please take a look at my comments.

packages/elements/src/components/theme-switch/theme-switch.scss
hugo-vrijswijk: I noticed some warnings from the sass compiler because its new version won't support the `/` syntax anymore. So I ran the migration script for it. I can revert the changes if you want

hugo-vrijswijk:
> Too bad this is a breaking change for typescript projects

What makes this a breaking change? All types can still be imported from the root, this just adds an extra export

nicojs:
Aha, that makes sense! But why did I spend an hour on it last Sunday then? 🤦‍♂️

ClemensSahs:
He please add `".": "./dist/src/index.js",` to the `exports`-key from `package.json` of the package `mutation-testing-report-schema`

```
"main": "dist/src/index.js",
  "typings": "dist/src/index.d.ts",
  "exports": {
    ".": "./dist/src/index.js", // <- this line
    "./api": "./dist/src-generated/schema.js"
  },```

Reason is that the `exports` feature will overwrite the `main` export so we don't have a real `main` and must defined as a duplicate.

thanks and best regards

nicojs:
Ahhh that's why I had the issue with migrating this in StrykerJS. Great find @ClemensSahs

PR #1126, opened 2021-05-22
feat(elements): upgrade to bootstrap 5

hugo-vrijswijk:
Update to Bootstrap 5. There are some small visual changes, but nothing big. Overall, I think the changes that are there are an improvement, but that's just opinion.

Screenshot diffs are probably the easiest way to see the differences


nicojs (approved):
Thanks for the work! Quick question: is this font color change expected? Or should it remain white?

image

hugo-vrijswijk:
> Thanks for the work! Quick question: is this font color change expected? Or should it remain white?

Fairly sure it's expected. In the [migration guide](https://getbootstrap.com/docs/5.0/migration/#sass) it's mentioned that the contrast values have been slightly changed. I'm guessing this is one of those changes

PR #1225, opened 2021-07-02
feat(metrics): add `aggregateResultsByModule` function

### LINKED ISSUE #1180: Add support for merging multiple reports together in 1 file
The stryker-dashboard currently supports merging of multiple report json files together into 1 (see [module](https://stryker-mutator.io/docs/General/dashboard#send-a-report-via-curl)).

https://github.com/stryker-mutator/stryker-dashboard/blob/eae4f8536eaef3daa89f48df4a00a4b831edefc3/packages/data-access/src/services/MutationTestingReportService.ts#L87-L95

However, I think it should be part of this `mutation-testing-metrics` package. That way others can also use it, see https://github.com/stryker-mutator/stryker-js/issues/2707

Also: Test results are currently not merged (not supported yet by the dashboard).

nicojs:
### `aggregateResultsByModule` [`(Record<string, MutationTestResult>) => MutationTestResult`]

Aggregates multiple reports together into a single report, grouped by module.

Input: `resultsByModule` The MutationTestResult objects by module name.

Closes #1180


packages/metrics/test/unit/aggregate.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+describe(aggregateResultsByModule.name, () => {
+  it('should result in an empty report when an empty object is provided', () => {
+    const expectedReport: MutationTestResult = {
+      files: {},
+      schemaVersion: '1.7',
+      thresholds: { high: 80, low: 60 },
+      config: {},
+      projectRoot: undefined,
+    };
+    expect(aggregateResultsByModule(Object.create(null))).deep.eq(expectedReport);
```
hugo-vrijswijk, line 15: What if an empty object is passed to the function? This `Object.create` thing makes me worry it gives unexpected results

hugo-vrijswijk (commented):
Some small comments about the code. But we've also talked about adding another possible field in the schema to allow multiple full reports aggregated inside a single schema (with different thresholds, configs, etc). How does that relate to this? I wouldn't want to merge this and expose it as a public API if we're going to make a very similar functionality soon as well

nicojs:
> But we've also talked about adding another possible field in the schema to allow multiple full reports aggregated inside a single schema (with different thresholds, configs, etc).

I still want that, but I don't want to spend the time for that a.t.m. I actually just want the merging of tests files in the dashboard, and wanted to move that out of the dashboard because users are asking for it as well.



> How does that relate to this? I wouldn't want to merge this and expose it as a public API if we're going to make a very similar functionality soon as well

When that functionality lands, I would expect the internals of this function to change so it provides the new format.

packages/metrics/src/aggregate.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param resultsByModule The MutationTestResult objects by module name
+ * @returns An aggregated result of all provided reports.
+ */
+export function aggregateResultsByModule(resultsByModule: Record<string, MutationTestResult>): MutationTestResult {
+  const projectRoots = Object.values(resultsByModule)
+    .map((report) => report.projectRoot)
+    .filter(isNotNullish);
+  const aggregatedResult: MutationTestResult = {
+    files: {},
+    schemaVersion: '1.7',
```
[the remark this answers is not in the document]
nicojs: I don't really know, but since we're also merging test views 🤷‍♀️

packages/metrics/test/unit/aggregate.spec.ts
nicojs: `Object.create(null)` is the emptiest object there is. Even more empty than `{}`, since it won't even have `Object.prototype` as its prototype

hugo-vrijswijk: Does `{}` give a different result?

packages/metrics/src/aggregate.ts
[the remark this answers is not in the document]
hugo-vrijswijk: It still works if you don't have a test view, right?

hugo-vrijswijk:
@nicojs What is the current status of this PR?

packages/metrics/test/unit/aggregate.spec.ts
nicojs: Of course 😌. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#custom_and_null_objects

hugo-vrijswijk: Then that should probably also have a test, right? `{}` seems a lot more common 🤔

packages/metrics/src/aggregate.ts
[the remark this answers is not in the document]
nicojs: Yes. So what would you suggest? 

We could also use `semver.max(...)` or equivalent.

A v1 schema also complies to v1.7, but not the other way around (ok, it might pass validation because backward compatible, but you get the point), that's why I chose that originally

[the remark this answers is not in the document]
hugo-vrijswijk: Ah, that makes sense. I'm just a little worried about problems a difference might cause. For example, validating a JSON that has `1.7` against the schema which has `1`. It also seems odd to me to only have this code-path result in a different schema version, while it doesn't create any JSON elements that weren't possible in e.g. `1.6`. 

Perhaps `semver.max` would be the best? But then we'd have to add it as a dependency, which might be a bit much. I guess since the schema is backwards compatible, `1` is still true 🤷‍♀️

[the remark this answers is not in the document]
nicojs: Semver won't work:

```
$ node -p "require('semver').lt('1.1', '1.2')"
C:\z\github\stryker-mutator\mutation-testing-elements\node_modules\semver\semver.js:332
    throw new TypeError('Invalid Version: ' + version)
    ^

TypeError: Invalid Version: 1.1
```

So I would personally like to keep it at 1.7. When we add more fields from later versions, we can up the version. I think it makes sense to keep this field updated.

Our JSON schema only validates that you're using `1.x`: `"pattern": "^1(\\.\\d*)?$"`, so it shouldn't pose an issue 🤷‍♀️

packages/metrics/test/unit/aggregate.spec.ts
nicojs: Good point. Added the test

nicojs:
I forgot about it 😶

hugo-vrijswijk (approved):

hugo-vrijswijk:
Ship it!

PR #1253, opened 2021-07-13
fix: package.json from package report-schema

ClemensSahs:
Node needs to defined a main in the exports, too.


hugo-vrijswijk:
Thanks for the fix @ClemensSahs!

PR #1426, opened 2021-10-07
test(chromedriver): revert chromedriver update

nicojs:


hugo-vrijswijk:
I think this fails because GH Actions has a set version of chrome installs, which now is 94. So unless we manually install the same version as the chromedriver this won't work

PR #1458, opened 2021-10-22
docs(mutators): add mutant operator name proposal

nicojs:
A proposal for mutant operator names. They are loosely based on the naming scheme found in infection PHP and pitest.

The goal of each mutant operator name is to have a conclusive name that describes the mutation. I used the following naming scheme:

`*NodeIdentification*[To*Replacement*]*Mutation*`

* `NodeIdentification`: Name the AST node being mutated.
* `Mutation`: Name the mutation being done:
  1. "Removal" when the node was removed.
  1. "Negation" when a node is replaced by its opposite ¹ 
  1. "Boundary" when a node is replaced by a similar node to test a boundary condition (like `<` to `<=`) 
  1. "Replacement" when a node is replaced by something other than the opposite or boundary.
* `Replacement` (only when the mutation was "Replacement"), name the node being replaced to ²

¹ I choose to interpret the "opposite" loosely here. When in doubt, I decided to use it. For example: "AdditionAssignmentNegation" for `+=` to `-=` and "AllMethodCallNegation" for `All` to `Any`.
² When naming the replacement node, I removed "expression", "operator" and "method call" for brevity. For example: "SingleOrDefaultMethodCallToFirstOrDefaultReplacement" and "NullishCoalescingOperatorToLogicalAndReplacement".

When identifying AST nodes, I did lean towards the JavaScript node types. Please verify if this is OK for C# and Scala, if not, please correct me. I used the following sources:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators
* https://sharplab.io/ 
* AST-explorer.net

We also have mutations that replace a part of a node, instead of the node itself. I had to invent logical names for them. They are summed up here:

* "FilledStringLiteralToEmptyReplacement" and friends
* "ObjectLiteralPropertiesRemoval"
* "ArrayConstructorItemsRemoval"
* "OptionalMemberExpressionOptionalRemoval"
* "OptionalComputedMemberExpressionOptionalRemoval"
* "OptionalCallExpressionOptionalRemoval"

I didn't do all of them, because I first want feedback on these names.


docs/supported-mutators.md
```diff
@@ -121,32 +123,32 @@ Stryker.NET _specific mutator_
...
-| Original                      | Mutated                       |
-| ----------------------------- | ----------------------------- |
-| `a && b`                      | <code>a &vert;&vert; b</code> |
-| <code>a &vert;&vert; b</code> | `a && b`                      |
-| `a ?? b`                      | `a && b`                      |
+| Mutant operator                                  | Original                      | Mutated                       |
+| ------------------------------------------------ | ----------------------------- | ----------------------------- |
+| LogicalAndOperatorNegation                       | `a && b`                      | <code>a &vert;&vert; b</code> |
+| LogicalOrOperatorNegation                        | <code>a &vert;&vert; b</code> | `a && b`                      |
+| NullishCoalescingOperatorToLogicalAndReplacement | `a ?? b`                      | `a && b`                      |
```
psfinaki, line 151: "Nullish" is not a concept in the .NET world, and I wouldn't know what to expect from this. Although I don't have a better idea, maybe only something even more verbose like "NullNullishCoalescingOperatorToLogicalAndReplacement".

(just passing by here - actually I think all this PR is a good idea)

rouke-broersma: I don't think Nullish makes sense anyway. The end result is that the value is treated in such a way that it is the same as Null, so NullCoalescing is good enough imo.

rouke-broersma:
I would propose to remove the syntax type and other classifiers (negation, operator, expression etc) where possible.

For example:

```
AdditionOperatorNegation -> Addition
AdditionAssignmentNegation -> AdditionAssignment
```

In the spirit of KISS, I think we should simplify where possible.

dvcopae:
Proposal has been extended in #2995.

nicojs:
I've changed the PR based on a conversation we just had. 

Attendees: @hugo-vrijswijk @rouke-broersma @richardwerkman @xandervedder 

- [x] Remove the technical node names from the node identifier. Like `MethodCall`, `Expression` and `Literal`.
    - This improves readability and makes it easier for non-technical people to understand (mind you, easier, not easy 😉).
- [x] Remove `Replacement` postfix. The `AbcToXyz` should be clear enough
- [x] Change `Nullish` in favor of `Null` for example: `NullCoalescingAssignmentToAndAssignment` instead of `NullishCoalescingAssignmentToAndAssignment`
- [x] ~`LogicalAnd`~ _`And`_. ~`LogicalOr`~ _`Or`_. ~`LogicalNot`~ _`Not`_. 
- [x]  `OptionalMemberToRequired` instead of `OptionalMemberRemoval` (my idea ☝)

docs/supported-mutators.md
```diff
@@ -293,32 +293,33 @@ Strings and literals identified to a regex are mutated in the following way:
...
 ## Unary Operator
 
-| Original | Mutated |
-| -------- | ------- |
-| `+a`     | `-a`    |
-| `-a`     | `+a`    |
+| Mutant operator | Original | Mutated |
+| --------------- | -------- | ------- |
+| PlusNegation    | `+a`     | `-a`    |
+| MinNegation     | `-a`     | `+a`    |
```
nicojs, line 312: Minus?

nicojs: It's minus.

docs/supported-mutators.md
```diff
@@ -293,32 +295,33 @@ Strings and literals identified to a regex are mutated in the following way:
...
-| -------------------------------------- | --------------------- |
-| `"foo"` (non-empty string)             | `""` (empty string)   |
-| `""` (empty string)                    | `"Stryker was here!"` |
-| `s"foo ${bar}"` (string interpolation) | `s""` ¹               |
+| Mutant operator                 | Original                               | Mutated               |
+| ------------------------------- | -------------------------------------- | --------------------- |
+| FilledStringToEmpty             | `"foo"` (filled string)                | `""` (empty string)   |
+| EmptyStringToFilled             | `""` (empty string)                    | `"Stryker was here!"` |
+| FilledInterpolatedStringToEmpty | `s"foo ${bar}"` (string interpolation) | `s""` ¹               |
+| FilledInterpolatedStringToEmpty | `\`foo ${bar}\``                       | `\`\``                |
```
hugo-vrijswijk, line 303: These lines break the website build: https://github.com/stryker-mutator/stryker-mutator.github.io/actions/runs/8295994425/job/22704167287#step:7:39

nicojs: Yep, sorry! https://github.com/stryker-mutator/mutation-testing-elements/pull/3066 fixes that.

Btw, it's not the line breaks, but the way of escaping backticks.

PR #1515, opened 2021-11-11
feat(diff): add new "diff" feature to mutant view 

### LINKED ISSUE #406: Feature request: Survived mutant stepper
It would be nice to be able to skip though a file by using some arrows.

Kinda like clicking next search result or next conflict in a merge confict editor. That way we can quickly check for survived and/or not covered mutants without scrolling trough the whole file.

Something like (excuse my bad designer skills):
image

This is especially useful when there are few survived mutants present in a large file. It's like _where's waldo_ at the moment 😅

### LINKED ISSUE #1419: Improve "diff" view
The current way we show the diff of a selected mutant is less then ideal. See this example:

It's great to see the diff in code, but I would also like to see an _actual_ diff. A la:

```diff
    this.options = args
-       .filter(arg => arg.substr(0, 1) === '-');
+       .filter(arg => true);
``` 

(but with code highlighting)

## Inspiration from other tools

Sonarqube uses an overlay.

image

[Infection's playground](https://infection-php.dev/r/kr1e)

## Technical details.

We're using [prismjs](https://prismjs.com/index.html) for code highlighting. It has something called the [diff highlighter plugin](https://prismjs.com/plugins/diff-highlight/) which looks awesome, but it [does not seem to be compatible](https://github.com/PrismJS/prism/issues/2393#issuecomment-630418463) with the [line number plugin](https://prismjs.com/plugins/line-numbers/) that we're also using. Maybe we can do something smart with showing the diff in an overlay? Not sure.

nicojs:
| old | new |
|-|-|
| image | image |

This adds the new "diff" feature to the mutant view. It also updated the test view to be in line with this way of working. The big changes:

1. New "diff" view when selecting a mutant. Still shown inline, but now with a clear "git diff" like feature, including `+`/`-` and background-color
  image
2. Since background color is used in the diff, we could no longer use that to mark mutants with a background color, instead using underline,
  image
3. To not interrupt the new diff view, we decided to move the markers of mutants to the _end of the line_. We also decided to change the look to be a "dot".
  image
4. Remove this underline when the mutant state is filtered out to form a calmer whole. 
  image
5. Since we can only show 1 diff at a time, we no longer support the "expand all" feature. Instead, we've introduced "next" and "previous" buttons at the top. This makes it easy to find the next surviving mutant, which closes #406.
6. Quality of life: it is also possible to click directly on the code when a mutant is underlined there. Clicking multiple times will allow you to "browse" through the mutants active at the cursor.

In action:

inline-diff-demo-10

Closes #1419


nicojs:
I think this PR is almost done. I only need to update the integration tests. Feel free to start the code review 🤓

CC @hugo-vrijswijk @simondel @rouke-broersma @richardwerkman

nicojs:
Green 💚🍏

nicojs:
@maks-rafalko you might also be interested in this

packages/elements/src/components/file/file.scss
```diff
@@ -1,17 +1,97 @@
...
+);
 
-@each $color, $value in $theme-colors {
-  .bg-#{$color}-light {
-    background-color: var(--bs-#{$color}-bg-light);
+// These 2 themes get underlined with a squiggly (wavy) for extra emphasis
+$mutant-squiggly: 'Survived', 'NoCoverage';
+
+@each $status, $theme in $mutant-themes {
+  @if (index($mutant-squiggly, $status) == null) {
```
hugo-vrijswijk, line 19: 🤯

packages/elements/src/components/file/file.scss
```diff
@@ -1,17 +1,97 @@
...
+@each $status in $mutant-squiggly {
+  .mte-selected-#{$status} .mutant.#{$status} {
+    /*
+  We want the underline of survived and no-coverage to be 'wavy'.
+  Don't use `text-decoration-style: wavy`, because chrome doesn't extend to all characters: https://stackoverflow.com/questions/57559588/how-to-make-the-wavy-underline-extend-cover-all-the-characters-in-chrome
+  Use border-image instead.
+  */
+    border-image-slice: 0 0 4 0;
+    border-image-width: 4px;
+    border-image-outset: 4px;
```
hugo-vrijswijk, line 46:
```suggestion
    border-image-outset: 3px;
```

packages/elements/src/components/test-file/test-file.component.ts
```diff
@@ -76,27 +108,52 @@ export class MutationTestReportTestFile extends LitElement {
...
-    if (code) {
-      highlightElement(code);
-
-      // Prism-js's `highlightElement` creates a copy of the DOM tree to do its magic.
-      // Now that the code is highlighted, we can bind the test components
-      this.forEachTestComponent((testComponent) => {
-        testComponent.test = this.model?.tests.find((test) => test.id === testComponent.getAttribute('test-id'));
-      });
-    }
+  private renderTests(tests: TestModel[] | undefined) {
```
hugo-vrijswijk, line 143: I think this method should be named differently, it's not rendering tests, it's rendering dots that mark a test. And maybe it should also be in a separate component (to be shared with source file component?)

package.json
```diff
@@ -14,7 +14,7 @@
     "prebuild": "npm run generate",
     "generate": "lerna run generate",
     "build": "tsc -b && lerna run build",
-    "test": "lerna run test --stream",
+    "test": "lerna run test --stream --concurrency 1",
```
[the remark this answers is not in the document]
nicojs: Probably not, will remove

packages/elements/src/components/file/file.component.ts
```diff
@@ -1,82 +1,168 @@
...
-    for (const mutantComponent of this.shadowRoot!.querySelectorAll('mte-mutant')) {
-      if (mutantComponent instanceof MutationTestReportMutantComponent) {
-        action(mutantComponent);
+  private codeClicked = (ev: MouseEvent) => {
+    ev.stopPropagation();
+
+    if (ev.target instanceof Element) {
+      let maybeMutantTarget: Element | null = ev.target;
+      const mutantsInScope: MutantModel[] = [];
+      for (; maybeMutantTarget instanceof Element; maybeMutantTarget = maybeMutantTarget.parentElement) {
```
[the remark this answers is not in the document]
nicojs: Only the first iteration. That's why I'm declaring it as `Element | null` at line 47, otherwise `maybeMutantTarget = mayMutantTarget.parentElement` won't work. Does that clear it up?

`parentElement` is of type `Element | null` (if I'm not mistaken).

packages/elements/src/components/file/file.scss
nicojs: sass > typescript

packages/elements/src/components/test-file/test-file.component.ts
nicojs: A separate component is a nice improvement, especially since we can then share it between mutant-dots and test-dots. For now, I'll just rename it if that's ok.

maks-rafalko:
This is very cool. I'm going to continue working on integration in https://github.com/infection/infection/pull/1149, already prioritized for Infection.

Thank you for working on this!

PR #1546, opened 2021-11-30
test(screenshots): update screenshots

hugo-vrijswijk:


nicojs:
So you think the lighter grey is better too? 🎓

PR #1547, opened 2021-11-30
docs(static mutants): add static mutants docs

nicojs:


docs/static-mutants.md
```diff
@@ -0,0 +1,70 @@
...
+  The static mutants will get a fresh test environment to run in. This might require a browser page refresh or the creation of a new process or app domain. Test filtering is limited since per test coverage cannot be determined. 
+
+## What Stryker does
+
+Different Stryker versions use different approaches here. In the table below, you can see which choices are available per framework.
+
+| Framework | Detect | Support run | Support ignore | Default |
+|--|--|--|--|--|
+| StrykerJS | ✅ | ✅ | ✅* | Run |
+| Stryker.NET| ❌ | ✅ | ❌ | Run |
```
hugo-vrijswijk, line 67: Is this meant to be Stryker.NET twice?

docs/static-mutants.md
```diff
@@ -0,0 +1,70 @@
...
++const hi = ''; 
+```
+
+All Stryker frameworks use mutant schemata, or mutation switching, to activate mutations during test executing. This means that the actual code Stryker produces looks more like this:
+
+```js
+const hi = global.activeMutant === '1' ? '' : '👋';
+```
+
+The goal is that Stryker would be able to activate this mutant during test execution by setting `global.activeMutant` to `'1'`. However, since this mutant is only executed once during startup. Activating or de-activating the mutant won't help. The `hi` const here is already declared, and activating mutant '1' doesn't change that.
```
psfinaki, line 43:
```suggestion
The goal is that Stryker would be able to activate this mutant during test execution by setting `global.activeMutant` to `'1'`. However, since this mutant is only executed once during startup, activating or de-activating the mutant won't help. The `hi` const here is already declared, and activating mutant '1' doesn't change that. 
```

docs/static-mutants.md
```diff
@@ -0,0 +1,70 @@
...
+You can identify static mutants in your report with the 🗿 static emoji.
+
+![static mutant](img/static-mutant.png)
+
+## How to deal with static mutants
+
+You can deal with static mutants in two ways:
+
+1. You can ignore them<br />
+   The mutants will be shown in your report with the "Ignored" state and won't count towards your mutation score. This feature is called `--ignoreStatic`
```
rouke-broersma, line 56: I prefer not to call features by what their commandline flag might be at this time.

docs/static-mutants.md
```diff
@@ -0,0 +1,70 @@
...
+
+## What Stryker does
+
+Different Stryker versions use different approaches here. In the table below, you can see which choices are available per framework.
+
+| Framework | Detect | Support run | Support ignore | Default |
+|--|--|--|--|--|
+| StrykerJS | ✅ | ✅ | ✅* | Run |
+| Stryker.NET| ❌ | ✅ | ❌ | Run |
+| Stryker.NET| ✅ | ❌ | ✅ | Ignore |
```
rouke-broersma, line 68: Stryker.NET detects static mutants and runs them, we do not currently support not running them.

docs/static-mutants.md
nicojs: No, second was Stryker4s.

docs/static-mutants.md
nicojs: Sure, will remove it.

docs/static-mutants.md
nicojs: Fixed

nicojs:
Thanks for the review guys. I've fixed your remarks. I also introduced a part about "hybrid mutants", please let me know what you think.

nicojs:
I've converted it to a draft. I've decided to wait until I have all ESM features in before releasing `--ignore-static`.

docs/static-mutants.md
```diff
@@ -0,0 +1,103 @@
...
+You can think of similar constructs in C# with [static constructors](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors) or in Scala with [singleton objects](https://docs.scala-lang.org/tour/singleton-objects.html#inner-main)
+
+## How to deal with static mutants
+
+You can deal with static mutants in two ways:
+
+1. You can ignore them<br />
+   The mutants will be shown in your report with the "Ignored" state and won't count towards your mutation score.
+2. You can run them<br />
+   The static mutants will get a fresh test environment to run in. This might require a browser page refresh or the creation of a new process or app domain. Test filtering is limited since per test coverage cannot be determined.
```
hugo-vrijswijk, line 60: Should we also mention this _likely_ has a big performance impact with most frameworks?

docs/static-mutants.md
```diff
@@ -0,0 +1,103 @@
...
+// greet.spec.js
+import { createGreeter } from './greet.js';
+
+it('should greet me', () => {
+  const greeter = createGreeter();
+  expect(greeter('me')).toBe('👋 me');
+});
+```
+
+As you can see, we've moved the `hi` constant to a factory method `createGreeter`. This method executes during unit testing in the "it should great me" test. However, the factory method also executes _during the loading of greet.js_.
```
hugo-vrijswijk, line 85:
```suggestion
We've moved the `hi` constant to a factory method `createGreeter`. This method is now executed **twice**. First _during the loading of greet.js_, and second during unit testing in the "it should great me" test. 
```

Language like 'as you can see' doesn't add anything. Also I think the text that it is executed twice in this example can be clearer.

nicojs: True

docs/static-mutants.md
nicojs: Good point

PR #1553, opened 2021-12-09
perf/improve rendering performance

### LINKED ISSUE #1478: Navigation is slow
When viewing a large report de navigation between pages is slow. I found this when viewing a report of rxjs. It takes about 4 seconds to switch from page as could be seen below.

## Example
slow_html_report

hugo-vrijswijk:
Fixes #1478

This PR has two commits to improve performance:

- perf(mutation-test-report-app): only recalculate metrics if report changed
  - Before, every time you navigate the `app` component would think there's an update to `report`, and re-calculate the metrics. This is a small change to only re-calculate the report when the report actually changes. If calculating the report is really slow, you'll now only notice it when opening the report the first time, instead of on every page navigation.
- perf(metrics): use Map for fast testId -> TestModel lookup
  - Relating tests to mutants' `coveredBy` and `killedBy` properties had a Big O of O(n^2) because for each mutant, it would loop through all the tests. The `relate` function in `calculateMetrics` also shows up in performance profiles as 99% of the time taken on a big report. Instead of the quadratic performance hit, `relate` now uses a `Map` to efficiently lookup a test to testIds from a mutant. When benchmarking with `console.time` on my machine in Edge with the large report in #1478, the `relate` function now takes about 180ms instead of 3.7 seconds, which makes opening the large report pretty much instant.
 
Before:
image

After:
image

### DIFF
--- a/packages/elements/src/components/app/app.component.ts
+++ b/packages/elements/src/components/app/app.component.ts
@@ -107,10 +107,14 @@ export class MutationTestReportAppComponent extends LitElement {
   }
 
   public async updated(changedProperties: PropertyValues) {
-    if ((changedProperties.has('path') || changedProperties.has('report')) && this.report) {
-      this.updateModel(this.report);
-      this.updateContext();
-      this.updateTitle();
+    if (this.report) {
+      if (changedProperties.has('report')) {
+        this.updateModel(this.report);
+      }
+      if (changedProperties.has('path') || changedProperties.has('report')) {
+        this.updateContext();
+        this.updateTitle();
+      }
     }
     if (changedProperties.has('src')) {
       await this.loadData();
--- a/packages/metrics/test/unit/calculateMetrics.spec.ts
+++ b/packages/metrics/test/unit/calculateMetrics.spec.ts
@@ -195,7 +195,7 @@ describe(calculateMutationTestMetrics.name, () => {
     expect(mutant1?.killedByTests).deep.eq([test2]);
     expect(mutant1?.coveredByTests).deep.eq([test1, test2, test3]);
     expect(mutant2?.killedByTests).undefined;
-    expect(mutant2?.coveredByTests).deep.eq([test1, test4]);
+    expect(mutant2?.coveredByTests).deep.eq([test4, test1]);
     expect(test1.coveredMutants).deep.eq([mutant1, mutant2]);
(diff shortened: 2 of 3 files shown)


packages/metrics/test/unit/calculateMetrics.spec.ts
```diff
@@ -195,7 +195,7 @@ describe(calculateMutationTestMetrics.name, () => {
     expect(mutant1?.killedByTests).deep.eq([test2]);
     expect(mutant1?.coveredByTests).deep.eq([test1, test2, test3]);
     expect(mutant2?.killedByTests).undefined;
-    expect(mutant2?.coveredByTests).deep.eq([test1, test4]);
+    expect(mutant2?.coveredByTests).deep.eq([test4, test1]);
```
hugo-vrijswijk, line 198: `coveredBy` on this mutant is actually `coveredBy: ['test-4', 'test-1']` so `test4` being first actually makes more sense to me here, I'm not sure why it was reversed beforehand

packages/elements/src/components/app/app.component.ts
```diff
@@ -107,10 +107,14 @@ export class MutationTestReportAppComponent extends LitElement {
   }
 
   public async updated(changedProperties: PropertyValues) {
-    if ((changedProperties.has('path') || changedProperties.has('report')) && this.report) {
-      this.updateModel(this.report);
-      this.updateContext();
-      this.updateTitle();
+    if (this.report) {
```
nicojs, line 110: Unrelated, but I think `update` is the better hook for this.

nicojs (approved):
😍 great improvements!

nicojs:
F.y.i. you've improved performance of `relate` from `O(n)=n²` to `O(n)=n` 🤯

hugo-vrijswijk:
> F.y.i. you've improved performance of `relate` from `O(n)=n²` to `O(n)=n` 🤯

Yeah whoever made that first version must've had a late night or something 😉

I guess you don't really notice it with all the test sets of <10 tests per mutant

nicojs:
> > F.y.i. you've improved performance of `relate` from `O(n)=n²` to `O(n)=n` 🤯
> 
> Yeah whoever made that first version must've had a late night or something 😉

Yeah, probably. Also, the reviewer. He/she must have had a really shitty day 💩

😉

PR #1757, opened 2022-03-02
docs(static-mutants): slightly better wording

nicojs:
Solving the race condition on the remarks of #1547 

@hugo-vrijswijk


nicojs:
As you can see, this PR has slightly better wording ;)

hugo-vrijswijk:
But slightly worse formatting 😱

hugo-vrijswijk (approved):

nicojs:
As you can see, much better formatting.

PR #1839, opened 2022-04-20
feat: improve accessibility of app and metrics-table components

hugo-vrijswijk:
Adding some aria descriptors, and making tabbing through the tabs a bit clearer (outline was not visible before)

### DIFF
--- a/packages/elements/src/components/metrics-table/metrics-table.component.ts
+++ b/packages/elements/src/components/metrics-table/metrics-table.component.ts
@@ -18,7 +18,7 @@ export interface Column<TMetric> {
   tooltip?: string;
   width?: TableWidth;
   category: ColumnCategory;
-  isHeader?: true;
+  isBold?: true;
 }
 
 @customElement('mte-metrics-table')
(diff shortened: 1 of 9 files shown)


packages/elements/src/components/metrics-table/metrics-table.component.ts
```diff
@@ -127,7 +132,7 @@ export class MutationTestReportTestMetricsTable<TFile, TMetric> extends LitEleme
         </td>
         <td style="width: 50px;" class="fw-bold text-center text-${coloringClass}">${valueIsPresent ? mutationScoreRounded : undefined}</td>`;
     }
-    return column.isHeader ? html`<th class="text-center">${value}</th>` : html`<td class="text-center">${value}</td>`;
+    return html`<td class="text-center ${column.isHeader ? 'fw-bold' : ''}" aria-describedby="${`tooltip-${column.key.toString()}`}">${value}</td>`;
```
nicojs, line 135: Why is this change needed? Is a `th` not desired here for some reason?

nicojs (approved):
Why didn't we need to update the screenshots? Didn't you improve the outline of the tabs?

packages/elements/src/components/metrics-table/metrics-table.component.ts
hugo-vrijswijk: Because it's not a table header 😁. It's very much a table data cell, just bold (this is the total number of mutants, for example, in a row). So it looks the same (bold if `isHeader`), but makes more sense semantically now

nicojs: Alright. So why do we keep the model name as `isHeader` then? 🤔

packages/elements/src/components/metrics-table/metrics-table.component.ts
```diff
@@ -18,7 +18,7 @@ export interface Column<TMetric> {
   tooltip?: string;
   width?: TableWidth;
   category: ColumnCategory;
-  isHeader?: true;
+  isBold?: true;
```
nicojs, line 21: This is a pretty `bold` statement!

nicojs (approved):

PR #1955, opened 2022-06-23
Changes \| to | on supported-mutators page.

bernarden:
Changes \\| to | on supported-mutators page:
https://stryker-mutator.io/docs/mutation-testing-elements/supported-mutators/


rouke-broersma:
Apparently prettier does not agree with something in your change. Could you run prettier on the docs?

PR #2044, opened 2022-08-19
feat(sonarqube): jq filter for import to SonarQube

apgrucza:
I was unsure of where best in this repository to put this jq filter, so feel free to change to a different location.

For context, see https://github.com/stryker-mutator/stryker-js/issues/3663


integrations/sonarqube/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Usage:
+
+```
+jq -f mutation-report-to-sonar.jq mutation.json > mutation-sonar.json
+```
+
+After `mutation-sonar.json` is generated, you can import it into SonarQube using the analysis parameter `sonar.externalIssuesReportPaths`. See [Generic Issue Import Format](https://docs.sonarqube.org/latest/analysis/generic-issue/) for more details.
+
+Please note that:
```
rouke-broersma, line 13: You should also add a note with:

"
There are a couple of limitations with importing external issues:

you can't manage them within SonarQube; for instance, there is no ability to mark them False Positive.
you can't manage the activation of the rules that raise these issues within SonarQube. External rules aren't visible on the Rules page or reflected in Quality Profiles.
External issues and the rules that raise them must be managed in the configuration of your linter.
"
See: https://docs.sonarqube.org/latest/analysis/generic-issue/

integrations/sonarqube/README.md
```diff
@@ -0,0 +1,17 @@
+# SonarQube integration
```
hugo-vrijswijk, line 1: This should probably be placed in the `docs/` folder, that way, it will also show up on the website :)

apgrucza: > This should probably be placed in the `docs/` folder, that way, it will also show up on the website :)

The `docs/` folder is specific to Stryker, and I was trying to keep this jq filter independent of Stryker as it can be used with reports from any mutation testing framework that uses the schema in this repository. But I've now moved it into the `docs/` folder anyway, as having this on the Stryker website would make it much more discoverable which is probably more important. Are any other changes required to get this published to the website?

hugo-vrijswijk (approved):
LGTM! @rouke-broersma if you have no other comments we can merge this :)

hugo-vrijswijk:
Now also live on the website :) https://stryker-mutator.io/docs/mutation-testing-elements/sonarqube-integration/

PR #2045, opened 2022-08-19
fix: Only render tabs if there are test files

### LINKED ISSUE #1782: Test files tab should be hidden when there are no testfiles
Currently the tests tab is hidden only if the testfiles key is missing. If it is present the tab is present, even if there actually are no files.

rouke-broersma:
Fix #1782 

Note that I did not manage to test the change as I could not figure out how to and there are no docs describing how to contribute.


hugo-vrijswijk:
Could you add a test case for this? I don't fully understand when this occurs

rouke-broersma:
@hugo-vrijswijk :D

hugo-vrijswijk:
I think the best place to fix this is in metrics. There's a check done there if `testFiles` is present, but there should probably be an extra check to not calculate test metrics if testFiles is an empty object:
image

image

Then no changes would be needed in `elements` because metrics would be undefined (which is already checked)

PR #2139, opened 2022-10-12
fix: don't create test metrics if test files is empty

### LINKED ISSUE #1782: Test files tab should be hidden when there are no testfiles

rouke-broersma:
fix #1782


rouke-broersma:
@hugo-vrijswijk better? :D

hugo-vrijswijk (approved):

PR #2140, opened 2022-10-12
Fix broken "Back to Top" buttons

kelvinwelter:
Hello guys! 

It seems like the "Back to Top" buttons are broken on the "Supported mutators" page. If you try to click it, nothing happens. So, I fixed the links on these buttons and replaced the relative path `#supported-mutators` with just `#`. This resolved the issue for me. 

Hope this PR help, guys!


hugo-vrijswijk:
Thanks! Looks like this was also broken on the website, but this should fix it in both places 🎉

hugo-vrijswijk (approved):

PR #2395, opened 2023-03-24
build: use nx for task caching

hugo-vrijswijk:
Since Lerna now has a lot of built-in stuff from Nx, this PR implements some niceties from it, while still keeping Lerna for core monorepo stuff

- Removes `lerna bootstrap` (deprecated, npm workspaces are now recommended)
- Setup nx caching
  - CI will now only run compile/test/etc if a PR changes that package/dependencies (or root). That means, for example, updating `lit` in `elements` will not run metrics-scala for no reason, this will speed up CI a lot
  - Updating a dependency of a package (like updating report-schema), will trigger a change in its dependencies (metrics and elements)
- Versioning is unchanged
- Devdependencies are now in the package that use it (mainly elements) to make better use of package-specific CI mentioned above
- No more 'waterfall' of script steps. Linting, building, testing is now done in the most optimal parallel way (probably my favourite feature)


nicojs:
>  Linting is still shared between packages. It'd be nice to have a split setup per package for this too

Why? What would we gain by it? I would like to have all eslint options shared.

packages/report-schema/package.json
```diff
@@ -28,5 +30,8 @@
   "bugs": {
     "url": "https://github.com/stryker-mutator/mutation-testing-elements/issues"
   },
-  "homepage": "https://github.com/stryker-mutator/mutation-testing-elements/tree/master/packages/elements#readme"
+  "homepage": "https://github.com/stryker-mutator/mutation-testing-elements/tree/master/packages/elements#readme",
+  "devDependencies": {
+    "ajv": "^8.6.3"
```
nicojs, line 35: Why is this lone `devDependency` needed here, instead of hoisting that to the top as well?

packages/elements/src/tsconfig.json
```diff
@@ -6,13 +6,5 @@
...
     "lib": ["es2020", "dom", "dom.iterable"]
-  },
-  "references": [
-    {
-      "path": "../../report-schema/src"
-    },
-    {
-      "path": "../../metrics/src"
-    }
-  ]
```
nicojs, line 17: Why aren't we using native project references anymore? Sounds like a step backwards.

package.json
```diff
@@ -4,18 +4,23 @@
   "version": "0.0.6",
   "description": "A suite of web components for a mutation testing report.",
   "main": "dist/mutation-test-elements.js",
+  "workspaces": [
+    "packages/*"
+  ],
+  "nx": {
```
nicojs, line 10: Why do we need config for `nx` in 2 places? Both `nx.json` and the `nx` key here.

nicojs (changes requested):
I'm not sure that we're doing the right thing here. It seems much more convoluted.

1. The monorepo config is scatered amongst multiple config files: `nx.json`, `lerna.json`, `nx` key in package.json and `workspaces` key in package.json.
2. `nx.json` is pretty big. I don't feel comfortable maintaining that.
3. We are reinventing the wheel with regard to compiling our sources. TypeScript has native support for project references, but we are choosing to use `nx` for this?
4. I don't now what the benefit is of the more complex setup.
5. If you want to get rid of `bootstrap`, isn't it easier to simply use workspaces and remove `lerna bootstrap` from postintsall?

package.json
hugo-vrijswijk: I've added a comment, but essentially this is to create a 'root' project. Because we have root tasks that need to run before submodule tasks can run (like schema:test depends on root:build)

packages/report-schema/package.json
hugo-vrijswijk: I've moved packages that are only used by a single module to that module. IMO it makes more sense that, for example, `lit` is a devDependency in elements instead of in root. Since report-schema is the only one that uses `ajv` it belongs in here.

I _think_ this should also give more finegrained behaviour when `nx` updates this package (it'll only run tasks for this package and its dependencies, instead of all of them)

hugo-vrijswijk (commented):
@nicojs  I've updated this PR to simplify the setup a little bit. TypeScript now uses project references again, and the `nx.json` is greatly simplified to only really know about what tasks depend on what. 

What that means is you don't get finegrained _task_ running (e.g. don't run `metrics` tests if only a test in `report-schema` has changed), but I agree the configuration is a bit too much to maintain. And it's an uncommon scenario anyway.

What we *do* still have with this is finegrainedness on a _package_ level. So if you make a change in `metrics-scala`, the other packages won't have to run their tests, for example. Or if you make a change in `elements`, then `metrics` will only build, and not test. This seems the more common scenario, and the config is fairly simple IMO.

packages/elements/src/tsconfig.json
hugo-vrijswijk: Agreed, so I reverted this

nicojs (approved):
I gave it a quick lookover on my phone from my vacation address. This looks much simpler indeed. Feel free to merge and please feel free to keep maintaining this 😉

PR #2425, opened 2023-04-04
feat(schema): add `Pending` status to JSON schema

### LINKED ISSUE #2424: Add `Pending` status to mutation-testing-report-schema.json
# Motivation

To give some background information: I am currently doing my internship at the company behind Stryker. The goal of my internship is to make the mutation testing report realtime, because currently the user has to wait before the complete mutation testing proces is done to see any feedback in the report.

To make this possible, a new status type should be added to the `status` property in the `mutation-testing-report-schema.json`. This new status will be called `Pending`. `Pending` means that the mutant is generated, but not yet run.

The reason why this is needed, is because of the following idea:
1. Every Stryker tool already knows what mutants are generated. This can be used to display the "empty" mutation report. This empty mutation report will be called the "initial state".
2. In the "empty" report, every mutant will have the state `Pending`.
3. To make the report realtime, we supply updates to the report from a Stryker tool using Server Sent Events. The only thing that would be sent is the mutant `id` and the new `state` of said mutant.
   - The reason for using Server Sent Events, is because the protocol is simpler and the communication direction is unidirectional. This makes it perfect for this usecase.
4. Using these events, everything will be updated in realtime in the report which will make it faster to recieve feedback<sup>[1]</sup>

<sup>1</sup> Note that the actual testing process will not be faster, that is up to the tool. This should only be for getting the actual feedback faster on-the-go.

xandervedder:
Do tell if more needs to be changed or tested.

closes #2424

BREAKING CHANGE: `Pending` is now also a valid mutant status

### DIFF
--- a/packages/elements/src/components/mutant-view/mutant-view.ts
+++ b/packages/elements/src/components/mutant-view/mutant-view.ts
@@ -118,7 +118,7 @@ const COLUMNS: Column<Metrics>[] = [
   {
     key: 'totalMutants',
     label: 'Total',
-    tooltip: 'All mutants (valid + invalid + ignored + pending)',
+    tooltip: 'All mutants (except runtimeErrors + compileErrors)',
     category: 'number',
     width: 'large',
     isBold: true,
--- a/packages/elements/src/components/mutant-view/mutant-view.ts
+++ b/packages/elements/src/components/mutant-view/mutant-view.ts
@@ -118,7 +118,7 @@ const COLUMNS: Column<Metrics>[] = [
   {
     key: 'totalMutants',
     label: 'Total',
-    tooltip: 'All mutants (valid + invalid + ignored)',
+    tooltip: 'All mutants (valid + invalid + ignored + pending)',
     category: 'number',
     width: 'large',
     isBold: true,
--- a/packages/elements/src/lib/html-helpers.ts
+++ b/packages/elements/src/lib/html-helpers.ts
@@ -43,7 +43,7 @@ export function getContextClassForStatus(status: MutantStatus) {
       return 'warning';
     case MutantStatus.Ignored:
     case MutantStatus.RuntimeError:
-    case MutantStatus.Pending: // For now
+    case MutantStatus.Pending:
     case MutantStatus.CompileError:
       return 'secondary';
   }
--- a/packages/elements/src/lib/html-helpers.ts
+++ b/packages/elements/src/lib/html-helpers.ts
@@ -82,9 +82,9 @@ export function getEmojiForStatus(status: MutantStatus) {
     case MutantStatus.Survived:
       return renderEmoji('👽', status);
     case MutantStatus.Timeout:
-      return renderEmoji('⌛', status);
-    case MutantStatus.Pending:
       return renderEmoji('⏰', status);
+    case MutantStatus.Pending:
+      return renderEmoji('⌛', status);
     case MutantStatus.RuntimeError:
     case MutantStatus.CompileError:
       return renderEmoji('💥', status);
--- a/packages/elements/src/lib/html-helpers.ts
+++ b/packages/elements/src/lib/html-helpers.ts
@@ -84,7 +84,7 @@ export function getEmojiForStatus(status: MutantStatus) {
     case MutantStatus.Timeout:
       return renderEmoji('⌛', status);
     case MutantStatus.Pending:
-      return renderEmoji('⏳', status);
+      return renderEmoji('⏰', status);
     case MutantStatus.RuntimeError:
     case MutantStatus.CompileError:
       return renderEmoji('💥', status);
--- a/packages/elements/src/lib/html-helpers.ts
+++ b/packages/elements/src/lib/html-helpers.ts
@@ -43,6 +43,7 @@ export function getContextClassForStatus(status: MutantStatus) {
       return 'warning';
     case MutantStatus.Ignored:
     case MutantStatus.RuntimeError:
+    case MutantStatus.Pending: // For now
     case MutantStatus.CompileError:
       return 'secondary';
   }
@@ -82,6 +83,8 @@ export function getEmojiForStatus(status: MutantStatus) {
       return renderEmoji('👽', status);
     case MutantStatus.Timeout:
       return renderEmoji('⌛', status);
+    case MutantStatus.Pending:
+      return renderEmoji('⏳', status);
     case MutantStatus.RuntimeError:
     case MutantStatus.CompileError:
       return renderEmoji('💥', status);
--- a/packages/metrics-scala/metrics/src/main/scala/mutationtesting/MetricsResult.scala
+++ b/packages/metrics-scala/metrics/src/main/scala/mutationtesting/MetricsResult.scala
@@ -1,6 +1,9 @@
 package mutationtesting
 
 sealed trait MetricsResult {
+  /** The total number of mutants that are pending, meaning that they have been generated but not yet run.
+    */
+  def pending: Int
 
   /** At least one test failed while this mutant was active. The mutant is killed. This is what you want, good job!
     */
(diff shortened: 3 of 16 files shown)


hugo-vrijswijk:
Could you also add the status to metrics-scala?

packages/elements/src/lib/html-helpers.ts
```diff
@@ -82,6 +83,8 @@ export function getEmojiForStatus(status: MutantStatus) {
       return renderEmoji('👽', status);
     case MutantStatus.Timeout:
       return renderEmoji('⌛', status);
+    case MutantStatus.Pending:
+      return renderEmoji('⏳', status);
```
nicojs, line 87: I think we should keep emojis unique. I had to zoom in to see that this emoji is different than the one for a timeout. Maybe we should change the timeout status emoji to: ⏰

Thoughts @hugo-vrijswijk, @rouke-broersma @richardwerkman  ?

nicojs (approved):
Seems like a good start. We should also update [mutant-states-and-metrics.md](https://github.com/stryker-mutator/mutation-testing-elements/blob/master/docs/mutant-states-and-metrics.md), but that can be done in a separate PR.

I have a small remark. I'm curious what others think.

nicojs:
@maks-rafalko f.y.i. we're adding a `pending` state to a mutant as we're adding support for real-time updates to the HTML report. It's technically a breaking change since mutation testing reports that use the status won't be valid according to the v1 schema.

maks-rafalko:
thanks for letting me know. We, at Infection, didn't use `Pending` status so I assume we are not affected by this BC.

Looking forward to this feature, to migrate to Infection.

packages/elements/src/lib/html-helpers.ts
hugo-vrijswijk: I asked AI, and it came up with 3 answers:

🕒 Hourglass with Flowing Sand: This emoji can indicate that something is in progress or waiting for a certain duration.
⏳ Hourglass Done: This emoji can imply that something is almost done or nearing completion.
⚠️ Warning Sign: This emoji can suggest that something is uncertain or risky and may require attention or caution.
You can choose the emoji that best suits your needs or preferences. 😊

I think I like 🕒 the best

nicojs (changes requested):

packages/elements/src/lib/html-helpers.ts
```diff
@@ -82,6 +83,8 @@ export function getEmojiForStatus(status: MutantStatus) {
       return renderEmoji('👽', status);
     case MutantStatus.Timeout:
       return renderEmoji('⌛', status);
+    case MutantStatus.Pending:
+      return renderEmoji('⏰', status);
```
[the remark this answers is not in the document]
xandervedder: Ah, woops

xandervedder:
@nicojs I'm not so sure if this is a breaking change, since I tested my (concept) implementation with Stryker.NET. It appears that the mutation report accepts a status that is not defined in the schema, in my case this was the `NotRun` status that Stryker.NET uses internally.

Please correct me if I'm wrong.

rouke-broersma:
@xandervedder if the html report is not actually validating the received data, and instead the validation is done in the dashboard, then this is not a breaking change for the webcomponents but it is a breaking change for the schema, since the schema does enforce the status as an enum with known values. Unfortunately I think that does make this a breaking change after all. The webcomponents don't have their own versioning strategy, they are coupled to the schema.

packages/elements/src/lib/html-helpers.ts
rouke-broersma: How about ❔

richardwerkman: I like ⏰ for timeout. And ⏳ for pending

rouke-broersma: Or perhaps no emoji and no presence in the filters at all? Are they actually relevant as a filter?

richardwerkman: I think they are, otherwise, they are not counted in the total number of mutants right?

rouke-broersma: Why would they not be counted just because they don't have an emoji and aren't used as a filter?

nicojs:
F.y.i. this versioning thing is just a way to communicate to the outside world. It doesn't actually matter for us. So don't feel 'bad' about a breaking change. It's fine ☺️

packages/elements/src/lib/html-helpers.ts
xandervedder: Should we still change the current emojis to something else?

xandervedder:
Is there anything else I need to change for this to go through? Happy to change anything if needed 😃

xandervedder:
I think it should because otherwise the total amount of mutants would "grow".

Also, should pending be added to the columns in the  `mte-metrics-table`?

rouke-broersma:
I don't think pending should be in the table because it's only interesting for a very short time, and we can better use the progress bar to display this temporary state.

packages/elements/src/lib/html-helpers.ts
rouke-broersma: @xandervedder let's create a poll or something with the different options so we can make a decision.

nicojs:
We can also document it in a separate PR IMO. Merge this?

packages/elements/src/lib/html-helpers.ts
xandervedder: I will make one in the Stryker slack (general)

xandervedder:
I'll make a poll in the Stryker slack (#general).

packages/metrics-scala/metrics/src/main/scala/mutationtesting/MetricsResult.scala
```diff
@@ -59,7 +62,7 @@ sealed trait MetricsResult {
 
   /** All mutants.
     */
-  lazy val totalMutants: Int = totalValid + totalInvalid + ignored
+  lazy val totalMutants: Int = totalValid + totalInvalid + ignored + pending
```
rouke-broersma, line 65: pending should probably be added to totalValid instead of to totalMutants directly

packages/elements/src/lib/html-helpers.ts
```diff
@@ -43,6 +43,7 @@ export function getContextClassForStatus(status: MutantStatus) {
       return 'warning';
     case MutantStatus.Ignored:
     case MutantStatus.RuntimeError:
+    case MutantStatus.Pending: // For now
```
rouke-broersma, line 46: What does this comment mean?

packages/elements/src/components/mutant-view/mutant-view.ts
```diff
@@ -118,7 +118,7 @@ const COLUMNS: Column<Metrics>[] = [
   {
     key: 'totalMutants',
     label: 'Total',
-    tooltip: 'All mutants (valid + invalid + ignored)',
+    tooltip: 'All mutants (valid + invalid + ignored + pending)',
```
rouke-broersma, line 121: I would say pending is already included in valid. Though I'm not sure why we mention valid + invalid + ignored since valid and invalid are not columns in this table.

```suggestion
    tooltip: 'All mutants (except runtimeErrors+ compileErrors)',
```

packages/metrics/src/calculateMetrics.ts
```diff
@@ -168,7 +170,7 @@ function countFileMetrics(fileResult: FileUnderTestModel[]): Metrics {
     totalValid,
     totalInvalid,
     mutationScore: totalValid > 0 ? (totalDetected / totalValid) * 100 : DEFAULT_SCORE,
-    totalMutants: totalValid + totalInvalid + ignored,
+    totalMutants: totalValid + totalInvalid + ignored + pending,
```
rouke-broersma, line 173: pending should probably be added to totalValid instead of to totalMutants directly

PR #2453, opened 2023-04-20
feat(elements): realtime reporting

### LINKED ISSUE #2434: Add support for realtime reporting in mutation-testing-elements
Depends on #2424 (#2425).

---

# Realtime Reporting

To make the realtime reporting available for all the different Stryker tools, the mutation report itself needs to support the incoming events that the different tools will send.

## Motivation

See #2424 for more information.

## Implementation

The implementation will be done by using the [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) class in the browser. The general idea is that the underlying data structure will be updated (the JSON of the report), after that the report will wait for a certain amount of time (50-150ms). When it's done it will re-render the report. We don't want to re-render every time an event is received as this could lead to performance problems.

The only problem with sending events to the browser, is that these are not persistent. Which means if the user refreshes the page, all these events will be lost. I might open a different issue for that, but for now, all Stryker tools should still write the complete version to disk when the mutation testing is complete.

## Things to consider

- [ ] Unit testing
- [ ] Mutation testing
- [ ] Code smells

### Performance testing
_Note: only tested on FireFox so far._

Sending events from a Stryker tool to the browser is very fast and not the issue here. The issue here is re-rendering and some optimizations had to be made to retrieve mutants faster. 

#### Concept implementation

The first thing I tried - to make it work - was to find the same mutant that was sent to the browser and modifying said mutant:

```js
const theMutant = Object.values(this.report.files) 
  .flatMap(file => file.mutants)
  .find(mutant => mutant.id === newMutantData.id);
```

At first glance this works fine, but when sending thousands of events per second, this turns out to be too [slow](https://profiler.firefox.com/public/dxsqfyzex7cfgnskxsjdby2kqvvn776gb66cj90/calltree/?globalTrackOrder=0w2&implementation=js&thread=2&v=8), likely because I was doing two loops per update (which causes the slowdown).

#### Faster implementation

To speed things op dramatically, we can use a dictionary with every mutant inside of it, using its id as the key.

First, we have to prepare a dictionary:
```ts
private prepareMutantDatastructure() {
  if (!this.report) {
    return;
  }

  const allMutants = Object.values(this.report.files)
      .flatMap(file => file.mutants);
  
  allMutants.forEach(mutant => {
    this.mutants[mutant.id] = mutant;
  });
}
```

After preparing, accessing it is easier and way [faster](https://profiler.firefox.com/public/dtmtjc8em4gs3v6w77xqjekba9g4rkegx2n27v0/calltree/?globalTrackOrder=0w2&implementation=js&thread=2&v=8):

```js
const theMutant = this.mutants[newMutantData.id];
```

#### Last thing to fix

Things are fast, but re-rendering on every incoming event is still causing slowdowns in the browser.

To fix this we can simply schedule a render, preferably every few ms or so:

```ts
private renderScheduled = false;
private updateTimeout = 1000 / 60;

private scheduleRender() {
  if (this.renderScheduled) {
    return;
  }

  this.renderScheduled = true;
  setTimeout(() => {
    if (!this.report) {
      return;
    }

    this.updateModel(this.report);
    this.updateContext();
    this.renderScheduled = false;
  }, this.updateTimeout);
}
```

I went for 16.xx ms, but lower is also possible.

xandervedder:
# Realtime reporting
With this PR the report now supports realtime reporting. This means that users no longer have to wait for the testing proces to complete, but can get updates while a mutation testing is in progress.

## Issues
This PR has some issues that could be resolved now or at a later time:

- [x] small: When opening the testresources, the realtime example does not work (because of an unknown port number).
- [x] small(?): Test should be related to mutants, so that these will show up in the drawer.
- [x] ~~medium: The entire report is updated now, we would like to only update what's necessary (this could be done later).~~ (later)
- [x] small: We should use a better type for the partial mutants, and probably do some validations (?).

## Note
~~This was tested using the implementation of Stryker.NET, although this does not support the test tab yet. This is something I need to test.~~

closes #2434


xandervedder:
Not sure why the tests were pending, will check tomorrow

packages/elements/test/integration/lib/SseServer.ts
```diff
@@ -0,0 +1,47 @@
...
+          'Cache-Control': 'no-cache',
+          Connection: 'keep-alive',
+          'Access-Control-Allow-Origin': '*',
+        });
+      }
+    });
+    this.server.listen();
+  }
+
+  public send(event: { name: string; data: object }) {
```
xandervedder, line 22: Something like this:
```suggestion
  public sendMutationResult(mutation: Mutation) {
```

packages/elements/testResources/realtime-reporting-example/index.html
```diff
@@ -0,0 +1,29 @@
...
+  <body>
+    <a href="/">Back</a>
+
+    <script>
+      const urlParams = new URLSearchParams(window.location.search);
+      const port = urlParams.get('port');
+      const report = document.createElement('mutation-test-report-app');
+      report.setAttribute('src', 'mutation-report.json');
+      report.setAttribute('sse', `http://localhost:${port}`);
+      document.body.appendChild(report);
```
xandervedder, line 17: This should still work when normally running `npm run start`

packages/elements/testResources/realtime-reporting-example/mutation-report.json
```diff
@@ -0,0 +1,287 @@
+{
+  "schemaVersion": "1",
+  "thresholds": {
+    "high": 80,
+    "low": 60
+  },
+  "projectRoot": "C:\\Users\\example\\Projects\\Wrappit\\Wrappit\\Wrappit",
```
xandervedder, line 7: I used a very small part of my own (open source) project here, is this okay?

packages/metrics/src/model/mutant-model.ts
```diff
@@ -12,6 +12,7 @@ function assertSourceFileDefined(sourceFile: FileUnderTestModel | undefined): as
  * Represent a model of a mutant that contains its test relationship
  */
 export class MutantModel implements MutantResult {
+  [key: string]: any;
```
xandervedder, line 15: This was necessary to fix compile errors, but is there a better way?

https://github.com/stryker-mutator/mutation-testing-elements/pull/2453/files#diff-0f3462bf2afd2ec63a1c6a56844b694f192bcba2798139a072110019706b4337R223-R225

xandervedder:
@nicojs I reverted the changes we made to the rootmodel yesterday, since we were doing things that were already being done in the `calculateMutationTestMetrics`. It made more sense to me to update the `report` property instead.

The changes made will also work for the coveredBy and killedBy properties now (which are also now updated when the drawer stays open :smile: ). This is not the best solution, but I feel like we should focus on reactivity in a different pr (I still want to figure out debounce, scheduledUpdates etc.).

packages/elements/test/integration/realtime-reporting.it.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    it('should update the state of a mutant', async () => {
+      server.on('client-connected', (c) => (client = c));
+      await page.navigateTo(`realtime-reporting-example/?port=${port}#mutant/Attributes/HandleAttribute.cs/`);
+      await page.whenFileReportLoaded();
+      expect((await page.mutantView.mutantDots()).length).to.equal(1);
+      const mutantPending = page.mutantView.mutantMarker('0');
+      expect(await mutantPending.underlineIsVisible()).to.be.true;
+
+      client.sendMutantTested(defaultEvent);
+      await sleep(25);
```
nicojs, line 51: Can we refactor so we don't need a magic sleep?

packages/elements/src/components/app/app.component.ts
```diff
@@ -136,6 +141,15 @@ export class MutationTestReportAppComponent extends LitElement {
   }
 
   private updateModel(report: MutationTestResult) {
+    if (!this.report) {
```
nicojs, line 144: Why does this method accept a `report` and at the same time uses `this.report`?

packages/elements/src/components/app/app.component.ts
```diff
@@ -177,9 +191,43 @@ export class MutationTestReportAppComponent extends LitElement {
...
+
   public connectedCallback() {
     super.connectedCallback();
     this.subscriptions.push(locationChange$.subscribe((path) => (this.path = path)));
+    this.initializeSSE();
+  }
+
+  private source: EventSource | undefined;
+
+  private initializeSSE() {
```
nicojs, line 203:
```suggestion
  private initializeSse() {
```

I think the casing should be `Sse`. Like `Xml`, `Html`.

packages/elements/src/components/app/app.component.ts
```diff
@@ -177,9 +191,43 @@ export class MutationTestReportAppComponent extends LitElement {
...
+      if (theMutant === undefined) {
+        return;
+      }
+
+      for (const [prop, val] of Object.entries(newMutantData)) {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        (theMutant as any)[prop] = val;
+      }
+
+      this.updateModel(this.report);
```
nicojs, line 225: How does this work? We've updated `theMutant`, how does it magically update the screen when we re-initialize `this.mutants` in `this.updateModel`?

packages/elements/package.json
```diff
@@ -11,7 +11,7 @@
     "test": "npm run test:unit && npm run test:integration",
     "test:unit": "karma start --single-run",
     "test:unit:debug": "karma start --browsers ChromeHeadlessDebug",
-    "test:integration": "mocha --forbid-only --config ./test/integration/.mocharc.jsonc",
+    "test:integration": "mocha --forbid-only --exit --config ./test/integration/.mocharc.jsonc",
```
nicojs, line 14: I think this `SseTestServer` should have a `close` method that closes the underlying HTTP server (if `listen()` was called). Use that in `afterEach`. That way you don't have to rely on the `--exit` (which is kind of a dirty hack).

nicojs (changes requested):
Looks great so far. I'm not a fan of the way reactivity is built in now. I would rather see a call to `requestUpdate()` instead of hacking in some solution.

packages/elements/test/integration/realtime-reporting.it.spec.ts
xandervedder: Hmm I'm thinking of a way to do this, but you have to do this since sending an event to the browser takes time (and the call is not blocking).. 🤔

packages/elements/src/components/app/app.component.ts
xandervedder: I think this is a leftover, but this could probably be removed and use the `report` parameter instead.

packages/elements/src/components/app/app.component.ts
xandervedder: It rebuilds the `this.rootModel` with the changed `this.report`, which includes all the new information. However this method does not trigger the re-render. The `this.updateContext` method triggers the repaint. I'm not sure if this triggers a full repaint or a partial one (I should investigate that...)

packages/elements/src/components/file/file.component.ts
```diff
@@ -211,18 +212,48 @@ export class FileComponent extends LitElement {
...
         this.toggleMutant(this.selectedMutant);
       }
     }
     super.update(changes);
   }
 
+  private dispatchEventIfMutantIsUpdated(): void {
+    const theSameMutant = this.model.mutants.find((mutant) => mutant.id === this.selectedMutant?.id);
+    // TODO: deep eq?
+    if (theSameMutant === undefined || this.selectedMutant?.status === theSameMutant.status) {
```
xandervedder, line 241: Should we use a package for this? Or should I manually do this check (since it is only for the mutant)?

packages/elements/package.json
xandervedder: Ah, I was wondering why they were still pending.. this explains a lot 😅

packages/metrics/src/model/metrics-result.ts
```diff
@@ -23,4 +29,68 @@ export interface MetricsResult<TFile = FileUnderTestModel, TMetrics = Metrics> {
...
+    this.metrics = metrics;
+    this.file = file;
+  }
+
+  public updateParent(value?: MetricsResult<TFile, TMetrics>) {
+    this.parent = value;
+    this.childResults.forEach((result) => result.updateParent(this));
+  }
+
+  public updateAllMetrics() {
```
nicojs, line 45: As I mentioned in one of our meetings: we shouldnt' need this method. In fact, we shouldn't need any public `calculate` method at all. Instead, we could move all this calculation magic inside the class and only recalculate on the fly when the metrics are requested and they are dirty (`undefined`).

So we start out with `#metrics: TMetrics | undefined` and in the getter of `metrics()` calculate when they are undefined. If we later update coverage or killed by from a child, we walk up the files and mark them dirty (setting `#metrics` to `undefined`).

What do you think?

nicojs (approved):
I had limited time to review this; feel free to merge this and pick my remarks up later. I would also like to review the whole once it is done.

packages/metrics/src/model/mutant-model.ts
```diff
@@ -32,6 +32,9 @@ export class MutantModel implements MutantResult {
   public killedByTests: TestModel[] | undefined;
   public sourceFile: FileUnderTestModel | undefined;
 
+  #_coveredByTests: Map<string, TestModel> = new Map();
```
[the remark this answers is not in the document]
xandervedder: I used an underscore as to not clash with the `coveredByTests` property. I didn't want to change that property to a map, since I wanted to keep the behavior the same.

But I agree that this could do without the `_`, so it might need a different name.

packages/metrics/src/model/metrics-result.ts
xandervedder: I assume that we would mark the `#metrics` undefined when we update the state of a mutant too? How does it work in that case?

Should we create a release with the tag `realtime-beta` when this gets merged?

rouke-broersma: If we feel confident enough to merge we should feel confident enough to release. No need for a beta release, we don't have a release process for beta's in mutation-testing elements anyway.

PR #2475, opened 2023-05-01
chore: add directory property in `packages.json`

### LINKED ISSUE #2474: Add `directory` to package json
Add `directory` to `repository` key in the package.json. That should help with renovate bot using a single PR. See for example: https://github.com/stryker-mutator/stryker-js/blob/78cfecce0cf3068c474ef675c7de21dacc87059b/packages/core/package.json#L18

xandervedder:
closes #2474

### DIFF
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -19,7 +19,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/stryker-mutator/mutation-testing-elements.git",
+    "url": "https://github.com/stryker-mutator/mutation-testing-elements.git",
     "directory": "packages/elements"
   },
   "license": "Apache-2.0",
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -19,7 +19,8 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/stryker-mutator/mutation-testing-elements.git"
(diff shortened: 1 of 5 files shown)


packages/elements/package.json
```diff
@@ -19,7 +19,8 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/stryker-mutator/mutation-testing-elements.git"
+    "url": "git+https://github.com/stryker-mutator/mutation-testing-elements.git",
```
nicojs, line 22: I think we should remove the`git+` prefix while we're at it. I'm pretty sure this isn't a valid URL, which might be a problem for renovate bot.

nicojs (changes requested):

packages/elements/package.json
xandervedder: I also see this in the root `package.json`, should I remove this one as well?

hugo-vrijswijk: Actually renovatebot doesn't look at `repository` much for grouping PR's. It's done hardcoded: https://github.com/renovatebot/renovate/pull/21905

nicojs: This blows my mind! 🤯

nicojs: This blows my mind! 🤯

nicojs (approved):

PR #2498, opened 2023-05-08
fix(elements): improve performance

### LINKED ISSUE #2496: realtime reporting: improve performance when recieving large amounts of mutant events
We can leverage the power of rxjs in this case, by using `throttleTime` for example.

xandervedder:
I'm having issues mocking the `rxjs` functions, any suggestions @nicojs?

---
Instead of recalculating every time an event comes through, only recalculate it during certain intervals. By using the `throttleTime` function from rxjs we can tell rxjs to ignore certain events during a certain time period.

before: https://profiler.firefox.com/public/m0578rpmtf3xxvq31wfwd5927jbsdgmtcgs3ekr/calltree/?globalTrackOrder=0w2&implementation=js&thread=2&v=9

after: https://profiler.firefox.com/public/fftmqbk76rwn26g2p9wwjzwwr6ef7b6gqh052g8/calltree/?globalTrackOrder=0w3&implementation=js&thread=5&v=9

closes #2496


packages/elements/src/components/app/app.component.ts
```diff
@@ -27,6 +27,9 @@ interface TestContext extends BaseContext {
   result?: MetricsResult<TestFileModel, TestMetrics>;
 }
 
+// How many milliseconds should be between updates
+const UPDATE_CYCLE_TIME = 100;
```
rouke-broersma, line 31: I think it would be useful to express this (as a comment) in the form of given x amount of mutants which currently takes y time to render (on 'a random' computer) so we can get a feeling of the performance characteristics. 

Like, given 5000 mutants on a single page that take 50ms to render we would have a margin of 50 percent between updates. Something like that, so we can judge if the cycle time should ever be changed up or down.

nicojs:
Why don't we implement fine-grained reactivity instead? Might be fast enough to handle big workloads without the need for throttling

xandervedder:
Right now, the function that costs the most performance is recalculating the metrics since that is done recursively for every item in the model (which could be a large model).

I remember you told me we could mark the mutants (?) as dirty so we would only have to update the marked ones. Not sure if that would improve performance and also not sure when we would update the marked ones.

This might not be the best solution, but it definitely improves performance 😃.

rouke-broersma:
> Why don't we implement fine-grained reactivity instead? Might be fast enough to handle big workloads without the need for throttling

That sounds like a good future improvement, but at this stage I think that might be a bit too much out of scope considering the other work that still needs to be done (real time report support in the dashboard for example). Could you accept this throttling solution in the intermediate time?

hugo-vrijswijk:
FYI the conventional commit type should be `perf`: https://github.com/angular/angular/blob/main/CONTRIBUTING.md#type

packages/elements/src/components/app/app.component.ts
xandervedder: Is this better?

rouke-broersma: Yes but I would still like to see a real world example of the time it actually takes for 1 render of x amount of mutants to take place so we can put the cycle time number into context.

xandervedder: I'm not sure if I can measure that, but rendering is probably instantaneous?

Previously there were 6k+ calls to the `update` method, now there are `3` (https://profiler.firefox.com/public/fftmqbk76rwn26g2p9wwjzwwr6ef7b6gqh052g8/calltree/?globalTrackOrder=0w3&implementation=js&range=4862m780&search=updateAllMe&thread=6&v=9)

xandervedder: Those three updates took 6ms, so 2ms for one update? (on my laptop)

packages/elements/src/components/app/app.component.ts
```diff
@@ -216,64 +225,81 @@ export class MutationTestReportAppComponent extends RealtimeElement {
...
       }
-
-      theMutant.update();
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-      theTest?.update();
-      mutantChanges.next();
     });
+
+    const applySubscription = fromEvent(this.source, 'mutant-tested')
+      .pipe(throttleTime(UPDATE_CYCLE_TIME))
```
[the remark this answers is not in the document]
xandervedder: Reading the documentation, I think you are right (and that I have been reading the docs wrong..).

xandervedder:
There seems to be a test on windows that is timing out, investigating that right now...

PR #2503, opened 2023-05-09
fix(elements-table): transition score colors

hugo-vrijswijk:
Looks nice with the real-time updates 😃

image

### DIFF
--- a/packages/elements/src/components/metrics-table/metrics-table.component.ts
+++ b/packages/elements/src/components/metrics-table/metrics-table.component.ts
@@ -144,7 +144,7 @@ export class MutationTestReportTestMetricsTable<TFile, TMetric> extends Realtime
           ${valueIsPresent
             ? html`<div class="h-3 w-full min-w-[24px] rounded-full bg-gray-300">
                 <div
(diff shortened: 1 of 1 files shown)


nicojs:
Do you have an example? I tried to run it locally with the "🚀 Realtime reporting example" but couldn't see the difference.

hugo-vrijswijk:
Yeah the realtime reporting example always pushes 'Killed' mutants, so it's pretty difficult to see as it's just about always green. You can see it better when switching directories (and the same line in the table has a different colour)

xandervedder:
> Do you have an example? I tried to run it locally with the "🚀 Realtime reporting example" but couldn't see the difference.
It might be nice if the example sends a random status 🤔

hugo-vrijswijk:
> > Do you have an example? I tried to run it locally with the "🚀 Realtime reporting example" but couldn't see the difference.
> 
> It might be nice if the example sends a random status 🤔

The issue is then that you approach a score of 50% 😃

rouke-broersma:
Looks nice!

PR #2505, opened 2023-05-09
docs(real-time): add information about implementing real-time reporting

### LINKED ISSUE #2504: Add documentation on how to integrate realtime reporting for other mutation testing frameworks
# Events

- `mutant-tested`
- `finished`

xandervedder:
closes #2504

### DIFF
--- a/docs/implementing-real-time-reporting.md
+++ b/docs/implementing-real-time-reporting.md
@@ -15,7 +15,12 @@ Currently, Stryker provides the report with updates using [Server Sent Events](h
 
 First things first, the mutation testing framework should be able to generate a report filled with mutants<sup>1</sup>. These mutants must have the state `Pending`.
 
-Once this report has been generated, it is recommended to automatically open this report for a better user experience.
+There are two ways of serving the generated report:
+
+1. Write the file to disk.
+2. Serve the report using a webserver.
+
+For both of these options we recommend opening the report automatically for a better user experience.
 
 _<sup>1</sup> This could change in the future. Instead, the only thing that would be necessary is to show the files without any mutants and add new mutants while they are being generated. This would allow the report to open earlier._
 
@@ -38,7 +43,7 @@ Connection: keep-alive
 Access-Control-Allow-Origin: *
 ```
 
-The `Access-Control-Allow-Origin: *` is necessary only if the report is opened from a file from disk. It is possible to be more strict with CORS if the report is hosted.
+The `Access-Control-Allow-Origin: *` is necessary only if the report is opened from a file from disk. It is possible to be more strict with CORS if the report is served with a webserver.
 
 When sending events to the browser, please note that the payload should look like the following:
 
@@ -96,7 +101,6 @@ data: null
 
 ## Limitations
 
-Please note that if a user refreshes the page while realtime reporting is ongoing, all the state will be lost. To keep this state, the following can be done:
+It is up to the implementer to keep the state. If the user refreshes the page, the report will have lost the state that was built up until that point.
 
-1. Keep track of every mutant event that has been tested. When a browser reconnects, send all previous mutant results.
-2. If the report is opened from a file, it is possible to write the report to disk once the mutation testing process is done. If the user refreshes the page, they will see the full report.
+To prevent lost data, the implementer has to keep track of all mutant events that have happened until that point. If a browser reconnects it is possible to send all those events to the browser using the `mutant-tested` event.
--- a/docs/implementing-real-time-reporting.md
+++ b/docs/implementing-real-time-reporting.md
@@ -1,0 +1,82 @@
+---
+title: Implementing real-time reporting
+custom_edit_url: https://github.com/stryker-mutator/mutation-testing-elements/edit/master/docs/implementing-real-time-reporting.md
+---
+
+Real-time reporting was recently implemented in Stryker. This page is created to help other mutation testing frameworks to make use the same functionality.
+
+## How does it work?
+
+Currently, Stryker provides the report with updates using [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events). This was chosen over WebSockets because the dataflow should only be from a mutation framework to the report. The report should not have to send anything to the mutation framework.
+
+## Implementation details
+
+### Preparing the report
+
+First things first, the mutation testing framework should be able to generate an empty report. This report has all information necessary except the tested state of a mutant. These mutants must have the state `Pending`.
+
+Once this report has been generated, it is recommended to automatically open this report for a better user experience.
+
+### Changes to the `<mutation-test-report-app>` component
+
+To inform the report that it is going to be provided with updates, set the new `sse` property with a URL to where the SSE endpoint is located.
+
+```html
+<mutation-test-report-app sse="localhost:8080/sse"></mutation-test-report-app>
+```
+
+### Server side implementation
+
+At the server side, several HTTP headers are necessary to let the browser know that it should keep the connection open:
+
+```http
+Content-Type: text/event-stream
+Cache-Control: no-cache
+Connection: keep-alive
+Access-Control-Allow-Origin: *
+```
+
+The `Access-Control-Allow-Origin: *` is needed, since the report is a file on disk.
+
+When sending events to the browser, please note that the payload should look like the following:
+
+```
+event: eventName\n
+data: "..."\n\n
+```
+
+For more information, refer to the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format).
+
+### Events
+
+The report currently listens to two events:
+
+- `mutant-tested`.
+- `finished`.
+
+#### The `mutant-tested` event
+
+This event should be sent when a single mutant has been tested. The mutant should conform to the report schema. Every single property sent will be changed in the report.
+
+For this event the `data` field in the event should be stringified JSON:
+
+```http
+event: mutant-tested
+data: "{\"id\":\"0\",\"status\":\"Killed\",\"coveredBy\":[\"test_1\"]}"
+```
+
+#### The `finished` event
+
+To notify that the report should stop listening for events, send this event:
+
+```http
+event: finished
+data: null
+```
+
+## Caveat
+
+Please note that if a user refreshes the page while realtime reporting is ongoing, all the state will be lost. We have no plans on doing anything to combat this. Instead, it is possible to do the following:
+
+1. Keep track of every mutant event that has been tested. When a client connects (browser), send said events.
+2. When mutation testing is complete, write the full report to disk again. This way the report will still be filled when the user refreshes the page.
--- a/docs/implementing-real-time-reporting.md
+++ b/docs/implementing-real-time-reporting.md
@@ -13,7 +13,7 @@ Currently, Stryker provides the report with updates using [Server Sent Events](h
 
 ### Preparing the report
 
-First things first, the mutation testing framework should be able to generate a report filled with mutants<sup>1</sup>. These mutants must have the state `Pending`.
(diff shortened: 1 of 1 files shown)


docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,82 @@
...
+
+## How does it work?
+
+Currently, Stryker provides the report with updates using [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events). This was chosen over WebSockets because the dataflow should only be from a mutation framework to the report. The report should not have to send anything to the mutation framework.
+
+## Implementation details
+
+### Preparing the report
+
+First things first, the mutation testing framework should be able to generate an empty report. This report has all information necessary except the tested state of a mutant. These mutants must have the state `Pending`.
```
rouke-broersma, line 16: Is this not optional? I don't see why but if a project doesn't want to generate this initial state the report still works right?

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,82 @@
...
+At the server side, several HTTP headers are necessary to let the browser know that it should keep the connection open:
+
+```http
+Content-Type: text/event-stream
+Cache-Control: no-cache
+Connection: keep-alive
+Access-Control-Allow-Origin: *
+```
+
+The `Access-Control-Allow-Origin: *` is needed, since the report is a file on disk.
```
rouke-broersma, line 39: This is how we chose to implement it. However this depends on how the other project chooses deploy the report, if they don't launch from a file they can implement cors more strictly.

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,82 @@
...
+### Events
+
+The report currently listens to two events:
+
+- `mutant-tested`.
+- `finished`.
+
+#### The `mutant-tested` event
+
+This event should be sent when a single mutant has been tested. The mutant should conform to the report schema. Every single property sent will be changed in the report.
```
rouke-broersma, line 59: Specify which part of the schema

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,82 @@
...
+```http
+event: finished
+data: null
+```
+
+## Caveat
+
+Please note that if a user refreshes the page while realtime reporting is ongoing, all the state will be lost. We have no plans on doing anything to combat this. Instead, it is possible to do the following:
+
+1. Keep track of every mutant event that has been tested. When a client connects (browser), send said events.
```
rouke-broersma, line 81: I would rather explain this as: resend all previous mutant results on browser reconnect

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,82 @@
...
+2. When mutation testing is complete, write the full report to disk again. This way the report will still be filled when the user refreshes the page.
```
rouke-broersma, line 82: Only applies if the report is saved to disk in the first place

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,82 @@
...
+To notify that the report should stop listening for events, send this event:
+
+```http
+event: finished
+data: null
+```
+
+## Caveat
+
+Please note that if a user refreshes the page while realtime reporting is ongoing, all the state will be lost. We have no plans on doing anything to combat this. Instead, it is possible to do the following:
```
rouke-broersma, line 79: This sounds like we don't want to do anything about this, instead of this is just a limitation of the implementation, nothing can be done about this.

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,82 @@
...
+#### The `finished` event
+
+To notify that the report should stop listening for events, send this event:
+
+```http
+event: finished
+data: null
+```
+
+## Caveat
```
rouke-broersma, line 77: Limitations

rouke-broersma (changes requested):

docs/implementing-real-time-reporting.md
xandervedder: I'm not sure how one could do this without the initial state...

docs/implementing-real-time-reporting.md
xandervedder: Should I include the schema or link to it?

docs/implementing-real-time-reporting.md
rouke-broersma: Doesn't the mutant update contain pretty much everything one needs to build the report?

The only initial state absolutely necessary is the files, right?

xandervedder: No, it is not possible to do this without including the mutants. When a `mutant-tested` event is sent, it will use the `id` property to find the already present mutant and update its properties. We cannot add new mutants to the files when sending events.

docs/implementing-real-time-reporting.md
rouke-broersma: A copy and a link are both fragile, rather link to the whole schema and mention the structure. Something like `files[].mutant`.

xandervedder: I forgot something here as well, I need to explain which properties are optional and which ones are required.

docs/implementing-real-time-reporting.md
rouke-broersma: Right, perhaps we could note that as a possible future feature?

xandervedder: Good idea :+1:

nicojs (approved):

docs/implementing-real-time-reporting.md
nicojs: This is a limitation that results from your choice to write an actual empty report to disk, rather than serving the report from the web server. If you serve it from your webserver yourself, you can make sure it has the up-to-date states already in place.

Also see my comment in the review.

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,102 @@
...
+event: finished
+data: null
+```
+
+## Limitations
+
+Please note that if a user refreshes the page while realtime reporting is ongoing, all the state will be lost. To keep this state, the following can be done:
+
+1. Keep track of every mutant event that has been tested. When a browser reconnects, send all previous mutant results.
+2. If the report is opened from a file, it is possible to write the report to disk once the mutation testing process is done. If the user refreshes the page, they will see the full report.
```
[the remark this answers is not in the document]
rouke-broersma: We thought about that but then you'd need to somehow redirect them to the file when the testrun is finished, or keep the webserver open.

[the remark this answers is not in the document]
xandervedder: The main issue with hosting the report is what you do when the reporting is done. Personally I don't think redirecting the user to a file is nice UX.

docs/implementing-real-time-reporting.md
rouke-broersma: Then you have the same issue but at the end of the run instead of during the run, because the webserver will be closed.

nicojs: Why would the webserver be closed? The mutation testing framework could keep it open, right?

I would keep these docs about MTE only and not make it more complex by focussing on the server side of things.

docs/implementing-real-time-reporting.md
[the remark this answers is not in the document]
nicojs: If we want we can have a discussion about UX after the report is done at a later time. For now, I would keep this out of the docs (as stated in [the discussion above](https://github.com/stryker-mutator/mutation-testing-elements/pull/2505#discussion_r1189508824))

docs/implementing-real-time-reporting.md
xandervedder: Right, so only specify what should be sent to the report and how that can be achieved, but that the hosting of the report (file or server) should be up to the framework.

hugo-vrijswijk: On the framework side that causes problems with _when_ to close it though. Do you keep the webserver open forever? Or close on 'enter'? What about on CI?

docs/implementing-real-time-reporting.md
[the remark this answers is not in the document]
xandervedder: > Whenever the report is opened, first push the 'base' state to SSE

I don't think this will work, as I talked about [here](https://github.com/stryker-mutator/mutation-testing-elements/pull/2505#discussion_r1189029301). 

But it could work in combination with the report property. It is possible to keep track of mutants that have been tested and serve those when a browser connects to the SSE endpoint (but it would require #2498 to be merged to be able to handle large amounts of events)

docs/implementing-real-time-reporting.md
rouke-broersma: @hugo-vrijswijk we have a progress bar/toast notification as a future improvement

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,102 @@
...
+
+## How does it work?
+
+Currently, Stryker provides the report with updates using [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events). This was chosen over WebSockets because the dataflow should only be from a mutation framework to the report. The report should not have to send anything to the mutation framework.
+
+## Implementation details
+
+### Preparing the report
+
+First things first, the mutation testing framework should be able to generate an empty report. This report has all information necessary except the tested state of a mutant<sup>1</sup>. These mutants must have the state `Pending`.
```
hugo-vrijswijk, line 16: This isn't an empty report. It's a report filled with mutants, all with the state 'pending'

docs/implementing-real-time-reporting.md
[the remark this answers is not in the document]
rouke-broersma: Let's mark filling the initial state using events instead of the report property as a future improvement. It's more or less the same feature as discussed here: https://github.com/stryker-mutator/mutation-testing-elements/pull/2505/files/1c96cae494a30ef9d1dfb1189901a2e56cb4f4c0#r1189030710

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,106 @@
...
+
+## How does it work?
+
+Currently, Stryker provides the report with updates using [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events). This was chosen over WebSockets because the dataflow should only be from a mutation framework to the report. The report should not have to send anything to the mutation framework.
+
+## Implementation details
+
+### Preparing the report
+
+First things first, the mutation testing framework should be able to generate a report filled with mutants<sup>1</sup>. These mutants must have the state `Pending`.
```
rouke-broersma, line 16:
```suggestion
First things first, the mutation testing framework should generate a report with an initial state of the mutants. If the mutants don't yet have a state because they have not been tested you can use the state `Pending`.
```

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,106 @@
...
+
+There are two ways of serving the generated report:
+
+1. Write the file to disk.
+2. Serve the report using a webserver.
+
+For both of these options we recommend opening the report automatically for a better user experience.
+
+_<sup>1</sup> This could change in the future. Instead, the only thing that would be necessary is to show the files without any mutants and add new mutants while they are being generated. This would allow the report to open earlier._
```
rouke-broersma, line 25:
```suggestion
```

This can imo only cause confusion. Let's leave future improvements to the issues, no need to mention that here.

docs/implementing-real-time-reporting.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Changes to the `<mutation-test-report-app>` component
+
+To inform the report that it is going to be provided with updates, set the new `sse` property with a URL to where the SSE endpoint is located.
```
rouke-broersma, line 29:
```suggestion
To inform the report that it is going to be provided with real-time updates, set the new `sse` property with the SSE endpoint URI.
```

PR #2513, opened 2023-05-11
docs(real-time): small updates to real-time reporter docs

nicojs:


nicojs:
Do we want these docs to be hosted on the stryker-mutator.io website? My guess is no, right?

xandervedder (approved):

rouke-broersma:
> Do we want these docs to be hosted on the stryker-mutator.io website? My guess is no, right?

I don't see why not? https://stryker-mutator.io/docs/mutation-testing-elements/implementing-real-time-reporting/

PR #2560, opened 2023-06-01
feat(elements): add progress bar for real-time reporting

### LINKED ISSUE #2551: Add progressbar to mutation-testing-elements
Currently the user has no idea whether or not the mutation testing process is done, we can fix that by adding a progress-bar.

xandervedder:
https://github.com/stryker-mutator/mutation-testing-elements/assets/34238405/5841ae3e-c233-4a3d-aa07-5e66e362e4aa

---
closes #2551


xandervedder:
Because I increased the padding (sticky bar in report), the snapshots are no longer correct. Will update these (if I figure out how)

hugo-vrijswijk:
> Because I increased the padding (sticky bar in report, to make room for the smaller progress-bar), the snapshots are no longer correct. Will update these (if I figure out how)

There's a manual workflow you can trigger for your branch to update the snapshots via GH Actions

packages/elements/src/components/metrics-table/metrics-table.component.ts
```diff
@@ -25,7 +25,7 @@ export interface Column<TMetric> {
 }
 
 @customElement('mte-metrics-table')
-export class MutationTestReportTestMetricsTable<TFile, TMetric> extends RealtimeElement {
+export class MutationTestReportTestMetricsTable<TFile, TMetric> extends RealTimeElement {
```
nicojs, line 28: It is abouttime to use `RealTime` 👍

packages/elements/src/components/progress-bar/progress-bar.component.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    this.#observer = new window.IntersectionObserver(([entry]) => {
+      if (entry.isIntersecting) {
+        this.#shouldBeSmall = false;
+      } else {
+        this.#shouldBeSmall = true;
+      }
+
+      this.requestUpdate();
+    });
```
nicojs, line 48: Could you add a comment here? What are you doing and why?

packages/elements/src/components/progress-bar/progress-bar.component.ts
```diff
@@ -0,0 +1,146 @@
...
+import { PropertyValues, html, nothing } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { tailwind } from '../../style';
+
+interface ProgressBarMetrics {
+  killed: number;
+  survived: number;
+  combined: number;
+  [key: string]: number;
+}
```
nicojs, line 12: See my previous remark. By removing this interface, the progress bar can be a lot simpler.

packages/elements/src/components/metrics-table/metrics-table.component.ts
hugo-vrijswijk: For `Real`

nicojs (changes requested):
Awesome addition. I love the visuals. Especially the way it moves to a tiny sliver of a bar.

Some small remarks.

packages/elements/src/components/app/app.component.ts
```diff
@@ -324,10 +324,11 @@ export class MutationTestReportAppComponent extends RealtimeElement {
         <div class="container bg-white pb-4 font-sans text-gray-800 motion-safe:transition-max-width">
           <div class="space-y-4 transition-colors">
             ${this.renderErrorMessage()}
-            <mte-theme-switch @theme-switch="${this.themeSwitch}" class="sticky top-offset z-20 float-right mx-4 pt-4" .theme="${this.theme}">
+            <mte-theme-switch @theme-switch="${this.themeSwitch}" class="sticky top-offset z-20 float-right pt-6" .theme="${this.theme}">
             </mte-theme-switch>
             ${this.renderTitle()} ${this.renderTabs()}
-            <mte-breadcrumb .view="${this.context.view}" class="my-4" .path="${this.context.path}"></mte-breadcrumb>
+            <mte-breadcrumb .view="${this.context.view}" .path="${this.context.path}"></mte-breadcrumb>
+            <mte-progress-bar .rootModel="${this.rootModel}" .visible="${this.sse}"></mte-progress-bar>
```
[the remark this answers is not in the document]
rouke-broersma: Actually I think it might be nice to have the bar visible when not in real time reporting mode as well.

packages/elements/src/components/progress-bar/progress-bar.component.ts
xandervedder: It's checking whether the progress-bar is visible or not (intersects with the `visible` window), but a comment doesn't hurt 😃

packages/elements/src/components/progress-bar/progress-bar.component.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  #calculatePercentage(metric: keyof ProgressBarMetrics) {
+    return this.total !== 0 ? (100 * this.metrics![metric]) / this.total : 0;
+  }
+
+  // TODO: keep?
+  // #renderTotalMutants() {
+  //   return html`<span class="absolute bottom-0 right-0 flex h-8 items-center pr-3 font-bold text-gray-800">${this.#total}</span>`;
+  // }
+}
```
xandervedder, line 146: Not sure if I should put this back in 🤔

psfinaki:
This will be awesome - thanks a lot!

packages/elements/src/components/app/app.component.ts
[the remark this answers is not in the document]
xandervedder: I'm also not sure if we should add all the other statusses (ignored, compileErrors, etc.), Right now they are combined

nicojs:
@rouke-broersma @psfinaki I had a discussion with @xandervedder about _always_ showing the progress bar. So also when the report is finished. In that case, it functions as a summary.

What do you think?

rouke-broersma:
> @rouke-broersma @psfinaki I had a discussion with @xandervedder about _always_ showing the progress bar. So also when the report is finished. In that case, it functions as a summary.
> 
> What do you think?

I actually suggested that here: https://github.com/stryker-mutator/mutation-testing-elements/pull/2560#discussion_r1214577310 :) 

So I approve

packages/elements/src/components/app/app.component.ts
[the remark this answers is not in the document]
rouke-broersma: Statusses should have the same colour in the progress bar as assigned to them in the theme, and they should be listed separately, perhaps add a tooltip to display the status name.

packages/elements/src/components/app/app.component.ts
```diff
@@ -324,10 +324,18 @@ export class MutationTestReportAppComponent extends RealtimeElement {
...
             </mte-theme-switch>
             ${this.renderTitle()} ${this.renderTabs()}
-            <mte-breadcrumb .view="${this.context.view}" class="my-4" .path="${this.context.path}"></mte-breadcrumb>
+            <mte-breadcrumb .view="${this.context.view}" .path="${this.context.path}"></mte-breadcrumb>
+            ${this.sse
+              ? html`<mte-progress-bar
+                  .killed="${this.rootModel?.systemUnderTestMetrics.metrics.killed}"
+                  .survived="${this.rootModel?.systemUnderTestMetrics.metrics.survived}"
+                  .pending="${this.rootModel?.systemUnderTestMetrics.metrics.pending}"
+                  .total="${this.rootModel?.systemUnderTestMetrics.metrics.totalMutants}"
```
nicojs, line 336: I think we're missing `NoCoverage` and `TimeOut`. Instead of using `killed` and `survived`, we should use `detected` and `undetected`. 

I'm also doubtful about the use of `totalMutants`, since they include `invalid` mutants. In practise, `totalMutants !== detected + undetected`. Won't that be a problem?

See https://stryker-mutator.io/docs/mutation-testing-elements/mutant-states-and-metrics/#metrics

nicojs (changes requested):
My second review. Also wondering if we should add two more screenshot tests. One for an expanded progress bar and one for a minimized progress bar.

packages/elements/src/components/progress-bar/progress-bar.component.ts
```diff
@@ -0,0 +1,126 @@
...
+  public survived = 0;
+
+  @property({ attribute: false })
+  public pending = 0;
+
+  @property({ attribute: false })
+  public total = 0;
+
+  #observer: IntersectionObserver | undefined;
+  #shouldBeSmall = false;
```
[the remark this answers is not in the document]
xandervedder: Do you mean `@state`? `@data` is not recognized for me 😕

packages/elements/src/components/app/app.component.ts
xandervedder: Should these also have different colors? Right now it has 4 different colors (green, yellow, red, grey)

packages/elements/src/components/app/app.component.ts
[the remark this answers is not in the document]
xandervedder: > Statusses should have the same colour in the progress bar as assigned to them in the theme
Check.
> perhaps add a tooltip to display the status name.
Already does :) (but it doesn't display the correct name yet)

packages/elements/src/components/app/app.component.ts
xandervedder: Do we not want invalid mutants in the progress bar?

psfinaki:
Yeah it can be always shown there, why not.

I agree with @ rouke-broersma that the bar itself could be more explanatory. Like, the most important is that things move there and the amount of green color increases. But at first I actually didn't entirely understand what the orange color meant - ignored mutants? mutants with unknown statuses?

Also, this depends on how we call it, but as a random user of random apps I am used to the "progress bars" that have the final thing clearly stated, as in - 5 tasks done out of 7, or 54% complete (out of 100%). Here, the total amount of mutants in the bar actually increases with time, which looks a bit strange. 

I mean, if you look at the report statistics and if you know how mutation testing works, yes it becomes clear, but there are opportunities to make this more intuitive :)

rouke-broersma:
@psfinaki  At the moment I don't think it has the pending mutant state yet which causes this problem. I think once all mutant states are added to the progress bar the progress bar should no longer change in amount of mutants.

psfinaki:
Oh I see. Well, yeah, that would be an improvement then.

packages/elements/src/components/progress-bar/progress-bar.component.ts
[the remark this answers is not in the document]
nicojs: ah yes, `@state` :)

nicojs (changes requested):

nicojs:
Maybe we should choose a different color for `Pending`. Right now we're using the `secondary` color, which we are also using for `CompileError`, `RuntimeError` and `Ignored`. If we use a different color then you can easily see the amount of work still to be done.

packages/elements/src/components/progress-bar/progress-bar.component.ts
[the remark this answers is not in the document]
xandervedder: Then we can't use `#` private, but I don't think that's an issue

xandervedder:
With pending in progress-bar:
https://github.com/stryker-mutator/mutation-testing-elements/assets/34238405/6652b13b-32c1-447e-8d41-487ffb3f41b6

psfinaki:
Yesss that looks great.

xandervedder:
Alright, I'll have to update all the snapshots since they all include the progress bar now :)

packages/elements/src/components/progress-bar/progress-bar.component.ts
rouke-broersma, line 1: result-status-bar is imo a better term since we also use this as an overall overview even when not using real time reporting.

packages/elements/src/components/app/app.component.ts
```diff
@@ -279,7 +279,7 @@ export class MutationTestReportAppComponent extends RealtimeElement {
     });
 
     const applySubscription = fromEvent(this.source, 'mutant-tested')
-      .pipe(debounceTime(UPDATE_CYCLE_TIME))
+      .pipe(auditTime(UPDATE_CYCLE_TIME))
```
xandervedder, line 282: I used `auditTime` here, but it might be better to use `sampleTime` here:

debouncetime-vs-throttletime-vs-audittime-vs-sampletime---rxjs-operators-compared(1)

xandervedder:
This pull request should be ready now.

packages/elements/src/components/app/app.component.ts
nicojs: That seems interesting. I don't feel strongly one way or the other. Let's create a separate PR for it if you think its better.

nicojs (approved):
Thanks for the fantastic work! Sorry that it took a while to respond and get it merged.

PR #2807, opened 2023-10-30
feat(svelte): support svelte syntax highlighting

nicojs:
Add svelte file type and syntax highlighting

image


packages/elements/package.json
```diff
@@ -53,6 +53,7 @@
     "mutation-testing-report-schema": "2.0.3",
     "postcss": "8.4.31",
     "postcss-loader": "7.3.3",
+    "prism-svelte": "^0.5.0",
```
hugo-vrijswijk, line 56: This package hasn't been updated in about 2 years. I'm not sure this is the way to go. Maybe we can get away with "good-enough" formatting by using html for svelte?

nicojs: Yes, I thought of that. However:

- `prism-svelte` does have 20k downloads / week. 
- Svelte syntax is pretty stable. So it might be a stable package because of that.
- I see some usage 'in the wild'. Some examples:
    - https://magidoc.js.org/svelte-plugins/prismjs
    - https://lihautan.com/notes/svelte-with-prism/#with-prism-svelte
- C'mon man, this looks awesome 🤩
     image

nicojs: vs screenshot of html formatting: 
image

xandervedder: It looks nice!

I don’t think there’s a better alternative (though I must admit, I didn’t search for that long). Personally I am okay with this 👌

nicojs: are you ok with this @hugo-vrijswijk ?

hugo-vrijswijk: Yeah it does look nice 😁

hugo-vrijswijk (approved):
Not sure where all these conflicts came from. But other than that I think it's fine to merge

PR #2825, opened 2023-11-06
build: overhaul build with vite & playwright

hugo-vrijswijk:
Overhauls the build (and tests) by setting up Vite to build all projects, and playwright for the elements integration tests. 

- Build on each project is now done with Vite. This means both CommonJS and ESM are published!
- Unit tests on elements use vitest (with browser mode). I couldn't get Vite to work with karma. Unfortunately, this also means no Stryker as it doesn't support browser mode (yet?). Browser setup uses playwright!
- Integration tests on elements are done with playwright. The page objects had to be overhauled a little bit. But they are much faster (20 seconds vs 2.5 minutes on my machine) and a lot less flaky.
  - This also means no more dependency on chromedriver, and thus no dependency on the version of chrome that GHA has 🥳
- Typescript is used for typechecking the project only, as vite handles building
-  All packages pass [publint](https://publint.dev/) and [arethetypeswrong](https://arethetypeswrong.github.io/)
- Dependencies of building each project _should_ all work when using nx (build on one project depends on the other)


nicojs:
This improvement is freakin' amazing 🤩! I haven't had time to look at it yet, but I will try to do it in a 1h train ride later today.

packages/elements/test/integration/lib/helpers.ts
```diff
@@ -1,20 +1,15 @@
 import { Page, TestInfo, expect, test } from '@playwright/test';
 import { isHeadless } from './browser';
 
-export function sleep(n = 1000) {
```
nicojs, line 46: But... without `sleep`, how will I be able to dream? 🥺

hugo-vrijswijk: No sleep for the wicked

nicojs (changes requested):
We've looked through the code together. 🤝 Some findings:

- Vite is now responsible for bundling each project. Only elements should be bundled. We should use plain `tsc --build` like before.
- It's OK to have a breaking change to move to `esm`. I didn't even realize this project was not ESM yet. 
- Dev experience: 
    - We should have an up-to-date CONTRIBUTING.md, including a cheat sheet.
    - Some commands should work from the root. Useful ones might be: `test` and `build`?
    - CTRL+SHIFT+B support and debugging (launch.json) should be updated

nicojs (approved):
This looks a lot better :). Just a few remarks:

- The task "Debug Unit Tests (elements)" is not working. How do we debug elements unit tests?
- The breaking change of not exporting `MutantStatus` anymore will hurt a bit in StrykerJS, as we use it all over the place (it was the only way to work with the mutant result). But I see why it is necessary.

I've updated the schema2ts script, so pull before changing anything else.

PR #2842, opened 2023-11-10
fix(result-status-bar): only use valid mutants

### LINKED ISSUE #2639: Progress bar colours don't match mutant status colour
image

Red is survived + nocoverage, however the reporting colour for nocoverage is orange:

image

Orange is ignored + runtime error + compile error, however the reporting colour for these statuses is grey

image

https://github.com/stryker-mutator/mutation-testing-elements/blob/master/packages/elements/src/components/result-status-bar/result-status-bar.ts#L117

xandervedder:
Only use valid mutants in the status bar.

---

fixes #2639


nicojs:
Awesome, thanks!

Issue #2925, opened 2023-12-21
Improve description/statusReason rendering


hugo-vrijswijk:
When the `description`/`statusReason` is filled, the report shows it when opening the mutant. Oftentimes this is filled with information from the test framework, such as an exception, diff or otherwise. A lot of frameworks fill this information with ansi-codes for colored text. For Stryker4s specifically, this will have a _list_ of exceptions, delimited by newlines. 

This means that a nicely reported error like this:

<img width="1462" alt="afbeelding" src="https://github.com/stryker-mutator/mutation-testing-elements/assets/10114577/b4620ecd-3e08-4a94-8154-90e55a08e4b4">

Ends up looking like this in the report:

<img width="1279" alt="afbeelding" src="https://github.com/stryker-mutator/mutation-testing-elements/assets/10114577/13707d2a-fffb-4c01-be1e-15312a120f63">

Which is a lot less readable.

I think we can do two things to fix this:

1. Render newlines (and tabs?) properly
2. Render ANSI codes. Something like [ansi_up](https://www.npmjs.com/package/ansi_up) could work, but it should be fairly easy to make something ourselves

PR #2926, opened 2023-12-21
fix(drawer): preserve whitespace when rendering mutant description and statusReason

hugo-vrijswijk:
Closes half of #2925 

Before:

afbeelding

After:

afbeelding


nicojs (approved):
Does this also remove the ansi coloring characters?

hugo-vrijswijk:
> Does this also remove the ansi coloring characters?

No, ansi coloring is a whole different beast. I quickly tested a library that is meant to output html from ansi, but it didn't support everything and requires use of unsafeHtml.

I think it should be possible to create something ourselves and output lit templateResults. Ansi has a system, you just need to understand it. That, and support the different escape characters per language (because of course)

PR #3023, opened 2024-02-19
feat(real-time): add real-time package

nicojs:
Add real-time package. It is a helper package to help with the nodejs side of a real-time report. It abstracts away all the nitty gritty server-send events stuff underneath.


tsconfig.settings.json
```diff
@@ -3,14 +3,13 @@
     "target": "es2022",
     "module": "node16",
     "moduleResolution": "node16",
+    "lib": ["es2022"],
```
hugo-vrijswijk, line 6: Isn't this already implied with `target: es2022`?

packages/real-time/test/unit/mutation-event-sender.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import { MutationEventSender } from '../../src/mutation-event-sender.js';
+import { expect } from 'chai';
+
+describe(MutationEventSender.name, () => {
+  let responseMock: sinon.SinonStubbedInstance<ServerResponse>;
+  let onCompleteStub: sinon.SinonStub<[], void>;
+
+  beforeEach(() => {
+    responseMock = sinon.createStubInstance(ServerResponse);
```
hugo-vrijswijk, line 14: It seems a bit strange to me to make a stub instance for something like a http response. Is it not possible to make the object ourselves and spy on it? Especially since the sut is doing quite a lot of side-effect calling on the response this doesn't give me faith that it is being used properly

.vscode/launch.json
```diff
@@ -10,14 +10,14 @@
...
-        "source-map-support/register",
         "--timeout",
         "999999",
         "--colors",
         "${workspaceFolder}/packages/report-schema/dist/test/**/*.js",
-        "${workspaceFolder}/packages/metrics/dist/test/**/*.js"
+        "${workspaceFolder}/packages/metrics/dist/test/**/*.js",
+        "${workspaceFolder}/packages/real-time/dist/test/**/*.js"
       ],
+      "runtimeArgs": ["--enable-source-maps"],
```
hugo-vrijswijk, line 20: 🎉

xandervedder (approved):
Looking forward to use this in `stryker-dashboard`! 🥳

tsconfig.settings.json
nicojs: No. `"target": "es2022"` implies `"lib": ["es2022", "dom", "dom.iterable"]`. 🤷‍♀️

packages/real-time/test/unit/mutation-event-sender.spec.ts
nicojs: It is a unit test. I want to avoid creating an _actual_ response. 

Do you mean to say it's better to write a custom fake that implements the `ServerResponse` interface?

nicojs:
@hugo-vrijswijk I've commented on both remarks you've made. Could you take a look and answer my questions?

tsconfig.settings.json
hugo-vrijswijk: Of course 🤷

packages/real-time/test/unit/mutation-event-sender.spec.ts
hugo-vrijswijk: Well, I expected `ServerResponse` to just be a data object carrying some stream objects or something. That way you could test the streams. If that's not the case, this might be the simplest approach. Then a custom fake doesn't have much advantage over a stub

hugo-vrijswijk: I clearly expected too much 🙃

nicojs: Haven't you learned by now? This is JavaScript after all 😉

hugo-vrijswijk (approved):
@nicojs not sure if you were waiting for specific approval from my side but this looks good 👍

PR #3068, opened 2024-03-15
refactor(real-time):  move update logic to metrics package

nicojs:
Make `mutation-testing-metrics` responsible for the real-time calculations in the report. This has a couple of advantages:

- Easier to reason about: the metrics are always up-to-date.
- Easier to maintain: no ad-hoc calculations are needed anymore
- Can be reused from StrykerJS itself, for example, in the console reporter

This change was already discussed here: https://github.com/stryker-mutator/mutation-testing-elements/pull/2453#discussion_r1178769871.

Note: this refactor is needed to enable fine-grained reactivity.


nicojs:
Yeah... still needs some more work

Issue #3151, opened 2024-04-27
Elements does not support rust highlighting


max-te:
I recently [attempted](https://github.com/max-te/mutant-report) to use `mutation-testing-elements` to display [`cargo-mutants`](https://mutants.rs/) results. (Related discussion [there](https://github.com/sourcefrog/cargo-mutants/discussions/281))

This entry in the schema lead me to believe I could just set the required `language` property to "rust", ensure that `prismjs/components/prism-rust` is loaded and get syntax highlighting:

https://github.com/stryker-mutator/mutation-testing-elements/blob/3b88bf7c702935744112812c8755dbbecae421e9/packages/report-schema/src/mutation-testing-report-schema.json#L25-L33

However, after a lot of debugging around prism itself, I found out that the `language` property is not passed to rust and instead the language is inferred from the file extension by [`determineLanguage`](https://github.com/stryker-mutator/mutation-testing-elements/blob/3b88bf7c702935744112812c8755dbbecae421e9/packages/elements/src/lib/code-helpers.ts#L30), which has a limited list of cases.

As it stands, to add the highlighting I had to patch `mutation-testing-elements` in one of two ways:
 - either extend `determineLanguage` with a new case,
 - or change `highlightCode` to accept a language instead of a file name and pass `this.model.language` in `FileComponent` [here](https://github.com/stryker-mutator/mutation-testing-elements/blob/3b88bf7c702935744112812c8755dbbecae421e9/packages/elements/src/components/file/file.component.ts#L214).

For my experiment I chose the second option. I think this is a better interface, as `language` is a mandatory field anyways and just loading the right `prismjs` package and setting the language is something the user of the component can do without having to modify `mutation-testing-elements` itself.

However, this did break a bunch of tests around `TestFile`s and maybe the file extension should be used as a fallback or alternative option. I can see there was related discussion in #1503, though I think that relates only to test files.

hugo-vrijswijk:
Hi! Each prism language highlighting plugin is loaded separately in https://github.com/stryker-mutator/mutation-testing-elements/blob/master/packages/elements/src/style/prism-plugins.ts. This is because we don't want to load in _all_ prism languages, as that would bloat the bundle considerably. 

[...]

PR #3181, opened 2024-05-12
feat(elements): add Rust syntax highlighting & icon

max-te:
As @hugo-vrijswijk suggested in #3151, this adds syntax highlighting and file icons for the Rust language. This will help integrating `cargo-mutants` output. The included example report.json is a result of my experimental conversion process.


packages/elements/src/lib/code-helpers.ts
```diff
@@ -54,6 +55,8 @@ export function determineLanguage(fileName: string): ProgrammingLanguage | undef
       return ProgrammingLanguage.gherkin;
     case 'svelte':
       return ProgrammingLanguage.svelte;
+    case 'rs':
```
hugo-vrijswijk, line 58: Could you add a case in `test/unit/lib/code-helpers.spec.ts` too? That will actually test the highlighting part of prism working

hugo-vrijswijk (approved):
~~Thanks! Do you think you could replace the dark-mode file icon with a lighter colour? In light mode it looks great, but in dark mode you can't really see the icon: ~~
image

Updated:
image

hugo-vrijswijk:
Thanks for this PR! I've released this in `3.1.0`

PR #3192, opened 2024-05-16
feat: adding dotnet metric report

phmonte:
### **Description:**

This is a draft for a metrics solution for dotnet.
The initial idea was to create a nuget package capable of interpreting a ` stryker-report.json` file and returning the corresponding metrics.

I took advantage and created a CLI with the aim of interpreting a file and returning it to the console in a structured way.

I would like your opinion on this implementation, I looked at the PR for scala and tried to do something similar.

I had questions about nomenclature and folder organization.

Just linking the initial [PR](https://github.com/stryker-mutator/stryker-net/pull/2903)

### DIFF
--- a/packages/metrics-dotnet/MutationMetrics/Metrics.cs
+++ b/packages/metrics-dotnet/MutationMetrics/Metrics.cs
@@ -1,0 +1,70 @@
+﻿using Mutation.Metrics.Core.Models;
+using System.Text.Json;
+
+namespace Mutation.Metrics.Core
+{
+    public static class Metrics
+    {
+        public static MetricResult GenerateScore(string fullPath)
+        {
+            try
+            {
+                double killed = 0;
+                double timeOut = 0;
+                double survived = 0;
+                double noCoverage = 0;
+                double runtimeErrors = 0;
+                double compileErrors = 0;
+                double ignored = 0;
+                double pending = 0;
+
+                using StreamReader reader = new(fullPath);
+                var reporterFile = reader.ReadToEnd();
+
+                JsonReport report = JsonSerializer.Deserialize<JsonReport>(reporterFile, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
+
+                foreach (var files in (report.Files))
+                {
+                    foreach (var mutants in files.Value.Mutants)
+                    {
+                        var status = Enum.Parse<MutantStatus>(mutants.Status);
+                        switch (status)
+                        {
+                            case MutantStatus.Killed:
+                                killed++;
+                                break;
+                            case MutantStatus.Timeout:
+                                timeOut++;
+                                break;
+                            case MutantStatus.Survived:
+                                survived++;
+                                break;
+                            case MutantStatus.NoCoverage:
+                                noCoverage++;
+                                break;
+                            case MutantStatus.CompileError:
+                                compileErrors++;
+                                break;                            
+                            case MutantStatus.Ignored:
+                                ignored++;
+                                break;
+                            case MutantStatus.Pending:
+                                pending++;
+                                break;
+                            case MutantStatus.RunTimeError:
+                                runtimeErrors++;
+                                break;
+                        }
+                    }
+                }
+                return new MetricResult(killed, timeOut, survived, noCoverage, runtimeErrors, compileErrors, ignored, pending);
+            }
+            catch (IOException e)
+            {
+                Console.WriteLine("The file could not be read:");
+                Console.WriteLine(e.Message);
+            }
+            return new MetricResult();
+        }
+    }
+}
--- a/packages/metrics-dotnet/MutationMetrics/Models/JsonReport.cs
+++ b/packages/metrics-dotnet/MutationMetrics/Models/JsonReport.cs
@@ -1,0 +1,19 @@
+﻿namespace Mutation.Metrics.Core.Models
+{
+    public struct JsonReport
+    {
+        public string SchemaVersion { get; set; }
+        public IDictionary<string, int> Thresholds { get; set; } 
+        public IDictionary<string, SourceFile> Files { get; set; }
+    }
+
+    public struct SourceFile
+    {
+        public ISet<JsonMutant> Mutants { get; set; }
+    }
+
+    public struct JsonMutant
+    {
+        public string Status { get; set; }
+    }
+}
--- a/packages/metrics-dotnet/MutationMetrics/Models/MetricResult.cs
+++ b/packages/metrics-dotnet/MutationMetrics/Models/MetricResult.cs
@@ -1,0 +1,27 @@
+﻿namespace Mutation.Metrics.Core.Models
+{
+    public struct MetricResult
+    {
+        public double Detected { get; private set; }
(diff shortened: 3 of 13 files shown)


rouke-broersma:
I think this implementation goes beyond the scope of what we want to provide with mutation testing elements, by providing a CLI that reads a json and prints the score to the console. 

I assume that this is the specific requirement you have for a package like this so I understand why you modeled it like this. However imo mutation testing elements should not go beyond the core of providing the programmatic schematic (model) of the report and the calculation methods that could be performed on this model. Any I/O should be provided by the consuming program.

I think a case can be made to provide json serialization like metrics-scala did, but even that might go too far.

@hugo-vrijswijk @nicojs since you both have existing implementations I would ask that you weigh in on the scope of packages we want to provide.

packages/metrics-dotnet/MutationMetrics/Mutation.Metrics.Core.csproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
rouke-broersma, line 4: Since this package should have zero dependencies I think we can target .NET Standard 2.0 without issue and achieve the highest level of compatibility.

packages/metrics-dotnet/MutationMetrics/Models/MetricResult.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public struct MetricResult
+    {
+        public double Detected { get; private set; }
+        public double Undetected { get; private set; }
+        public double Covered { get; private set; }
+        public double Valid { get; private set; }
+        public double Invalid { get; private set; }
+        public double TotalMutants { get; private set; }
+        public double MutatioScore { get; private set; }
```
rouke-broersma, line 11:
```suggestion
        public double MutationScore { get; private set; }
```

packages/metrics-dotnet/MutationMetrics/Models/MetricResult.cs
```diff
@@ -0,0 +1,27 @@
...
+        public double MutatioScoreBasedOnCoveredCode { get; private set; }
```
rouke-broersma, line 12:
```suggestion
        public double MutationScoreBasedOnCoveredCode { get; private set; }
```

packages/metrics-dotnet/MutationMetrics/Models/JsonReport.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Mutation.Metrics.Core.Models
+{
+    public struct JsonReport
```
rouke-broersma, line 3:
```suggestion
    public struct MutationTestResult
```

Since this is an implementation of the schema we should try to follow the schema as closely as possible:

https://github.com/stryker-mutator/mutation-testing-elements/blob/master/packages/report-schema/src/mutation-testing-report-schema.json

packages/metrics-dotnet/MutationMetrics/Models/JsonReport.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public string SchemaVersion { get; set; }
+        public IDictionary<string, int> Thresholds { get; set; } 
+        public IDictionary<string, SourceFile> Files { get; set; }
+    }
+
+    public struct SourceFile
```
rouke-broersma, line 10:
```suggestion
    public struct FileResult
```

packages/metrics-dotnet/MutationMetrics/Models/JsonReport.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public ISet<JsonMutant> Mutants { get; set; }
+    }
+
+    public struct JsonMutant
```
rouke-broersma, line 15:
```suggestion
    public struct MutantResult
```

packages/metrics-dotnet/MutationMetrics/Models/JsonReport.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Mutation.Metrics.Core.Models
+{
+    public struct JsonReport
+    {
+        public string SchemaVersion { get; set; }
+        public IDictionary<string, int> Thresholds { get; set; } 
+        public IDictionary<string, SourceFile> Files { get; set; }
```
rouke-broersma, line 7: I would like to see TestFiles added as well:

https://github.com/stryker-mutator/mutation-testing-elements/blob/master/packages/report-schema/src/mutation-testing-report-schema.json#L113

rouke-broersma, line 7: I would like to see projectRoot added as well:

https://github.com/stryker-mutator/mutation-testing-elements/blob/master/packages/report-schema/src/mutation-testing-report-schema.json#L174

rouke-broersma, line 7: I would like to see performance added as well:

https://github.com/stryker-mutator/mutation-testing-elements/blob/master/packages/report-schema/src/mutation-testing-report-schema.json#L179

packages/metrics-dotnet/MutationMetrics/Metrics.cs
```diff
@@ -0,0 +1,70 @@
...
+                double timeOut = 0;
+                double survived = 0;
+                double noCoverage = 0;
+                double runtimeErrors = 0;
+                double compileErrors = 0;
+                double ignored = 0;
+                double pending = 0;
+
+                using StreamReader reader = new(fullPath);
+                var reporterFile = reader.ReadToEnd();
```
rouke-broersma, line 22: This should be done by the CLI

packages/metrics-dotnet/MutationMetrics/Metrics.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                JsonReport report = JsonSerializer.Deserialize<JsonReport>(reporterFile, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
```
rouke-broersma, line 24: I'm not sure this belongs here, CalculateMetrics should probably receive a concrete class. We could perhaps provide a helper class for deserializing depending on the conversation on if this would be within scope for this package.

hugo-vrijswijk:
> since you both have existing implementations I would ask that you weigh in on the scope of packages we want to provide.

The main reason for the metrics packages is to have a single place per language that implements the [metrics interface].(https://stryker-mutator.io/docs/mutation-testing-elements/mutant-states-and-metrics/). IMO this is at _least_ having a model for a result and being able to (correctly) calculate metrics based on the result. And each language might add features that are useful for it. Like aggregating reports for JS, or a type-safe JSON codec for Scala.

Non-Stryker programs probably get the most use out of the first part to calculate score or other metrics from a report. I agree that reading JSON from disk or a CLI that writes to stdout is maybe _too_ specific for a package here

PR #3367, opened 2024-08-20
fix(elements): improve usage as a library

hugo-vrijswijk:
Contains 2 fixes:

- Add checks for `isServer` to prevent browser-specific code from running on the server. [This helps for SSR](https://lit.dev/docs/ssr/authoring/#browser-only-code).
- Externalize lit when building as a library (the default package export), and bundle it when building as an IIFE (for script usage on the page, when importing from `/dist/mutation-test-elements.js`). [Lit recommends externalizing when building as a library](https://lit.dev/docs/ssr/authoring/#bundlers).


packages/elements/vite.config.ts
```diff
@@ -31,6 +31,11 @@ export default defineConfig(
       },
       esbuild: esbuildOptions,
       build: {
+        emptyOutDir: !process.env.IIFE_BUILD,
+        rollupOptions: {
+          // For library usage, we want to externalize lit. When included as a script on the page we want to bundle it.
+          external: process.env.IIFE_BUILD ? undefined : [/^lit/],
```
hugo-vrijswijk, line 37: With this, lit _should_ be a dependency when using the root import, but I don't want to add lit as a dependency, for the use case where Stryker-js imports the packge and only needs the `mutation-test-elements.js` file. I'm not sure how this could be done. Maybe an `optionalDependencies`? @nicojs

packages/elements/src/lib/browser.ts
```diff
@@ -1,7 +1,10 @@
+import { isServer } from 'lit';
+
 /**
  * Test if localStorage exists and is enabled
  */
 export function isLocalStorageAvailable() {
+  if (isServer) return false;
```
hugo-vrijswijk, line 7: Because this is a static boolean based on [export conditionals](https://nodejs.org/api/packages.html#conditional-exports) this is really hard to test :(. Because our tests run in the browser this will always just be `false`

PR #3475, opened 2024-10-19
feat(elements): make mutant dots styling more clear

### LINKED ISSUE #3464: Feature: Change circle selection button color
The button in question:
image

When I select one of the buttons it's a bit confusing that it turns completely blue.
I suggest only adding a blue outline instead of making it completely blue.

image
image

Adel-Atzouza:
closes #3464


packages/elements/src/components/file/file.component.ts
```diff
@@ -134,15 +134,17 @@ export class FileComponent extends RealTimeElement {
...
-            svg`<svg mutant-id="${mutant.id}" class="mutant-dot ${
-              this.selectedMutant?.id === mutant.id ? 'selected' : mutant.status
-            }" height="10" width="12">
-          <title>${title(mutant)}</title>
-          <circle cx="5" cy="5" r="5" />
-          </svg>`,
-        )
+      ? mutants.map((mutant) => {
+          const selected = this.selectedMutant?.id === mutant.id;
+          const selectedClass = selected ? 'selected' : '';
```
hugo-vrijswijk, line 139: I don't think the selected class is used anymore. So it can be removed, right?

packages/elements/src/components/file/file.component.ts
```diff
@@ -134,15 +134,17 @@ export class FileComponent extends RealTimeElement {
...
+
+          return svg`
+            <svg mutant-id="${mutant.id}" class="mutant-dot ${selectedClass} ${mutant.status}" height="10" width="12">
+              <title>${title(mutant)}</title>
+              
+              ${selected ? svg`<polygon class="mutant-dot-polygon" points="5,10 0,0 10,0"  />` : svg`<circle cx="5" cy="5" r="5" color="red" />`}
```
hugo-vrijswijk, line 145:
```suggestion
              ${selected ? svg`<polygon class="mutant-dot-polygon" points="5,10 0,0 10,0"  />` : svg`<circle cx="5" cy="5" r="5" />`}
```

We don't want all mutant dots to be red

packages/elements/src/components/file/file.scss
```diff
@@ -31,8 +31,8 @@ $mutant-squiggly: 'Survived', 'NoCoverage';
   }
 }
 
-svg.mutant-dot.selected {
-  @apply fill-sky-400;
+polygon.mutant-dot-polygon {
+  @apply stroke-gray-800;
```
hugo-vrijswijk, line 35: For single tailwind classes I think it makes more sense to inline it. I.e. instead of the `polygon.mutant-dot-polygon` selector, add the `stroke-gray-800` class to the polygon directly. Then this bit can be removed

hugo-vrijswijk (changes requested):
A few minor comments, but this mostly looks good! 

Some of the tests will fail right now because they are doing screenshot comparisons against a previous version. The screenshots are taken on CI/CD so they always look the same (no different fonts etc). If you address the comments I can run the CI job to update the screenshots for you

hugo-vrijswijk (approved):

hugo-vrijswijk:
@Adel-Atzouza  thanks again for the contribution!

PR #3517, opened 2024-11-21
feat(file-picker): add fuzzy search to file picker

hugo-vrijswijk:
Also restyles the file picker list slightly and improves accessibility.

https://github.com/user-attachments/assets/6d453a3a-1b66-4dd7-a93f-fd4e1210cdfd


packages/elements/src/components/file-picker/file-picker.component.ts
```diff
@@ -232,49 +256,44 @@ export class MutationTestReportFilePickerComponent extends LitElement {
...
+  };
 
   #filter(filterKey: string) {
-    this.filteredFiles = Array.from(this.#searchMap.keys())
-      .filter((file) => file.toLowerCase().includes(filterKey.toLowerCase()))
-      .map((file) => ({ name: file, file: this.#searchMap.get(file)! }));
+    if (!filterKey) {
+      this.filteredFiles = this.#searchTargets;
+    } else {
+      this.filteredFiles = fuzzysort.go(filterKey, this.#searchTargets, { key: 'prepared', threshold: 0.3, limit: 500 }).map((result) => ({
```
xandervedder, line 289: Question: the limit here indicates the time? or is it something else?

xandervedder (approved):

packages/elements/src/components/file-picker/file-picker.component.ts
hugo-vrijswijk: It's the number of search results. Not super common to have > 500 files, but it is recommended to keep the search fast. Search was always around 1ms for me in the large example

PR #3679, opened 2025-03-14
docs: Add new Stryker.NET mutations

martincostello:
Add new unsigned right sift and Append/Prepend mutations for Stryker.NET.

See https://github.com/stryker-mutator/stryker-net/pull/3200 and https://github.com/stryker-mutator/stryker-net/pull/3199.

### DIFF
--- a/docs/supported-mutators.md
+++ b/docs/supported-mutators.md
@@ -56,13 +56,15 @@ All Stryker versions support a variety of different mutators. We've aligned on a
 | MultiplicationAssignmentNegation              | `*=`                 | `/=`                 |
 | DivisionAssignmentNegation                    | `/=`                 | `*=`                 |
 | RemainderAssignmentToMultiplicationAssignment | `%=`                 | `*=`                 |
-| LeftShiftAssignmentNegation                   | `<<=`                | `>>=`                |
-| RightShiftAssignmentNegation                  | `>>=`                | `<<=`                |
+| LeftShiftAssignmentNegation                   | `<<=`                | `>>=`, `>>>=`²       |
+| RightShiftAssignmentNegation                  | `>>=`                | `<<=`, `>>>=`²       |
+| UnsignedRightShiftAssignmentNegation²         | `>>>=`               | `<<=`, `>>=`         |
 | BitwiseAndAssignmentNegation                  | `&=`                 | <code>&#124;=</code> |
 | BitwiseOrAssignmentNegation                   | <code>&#124;=</code> | `&=`                 |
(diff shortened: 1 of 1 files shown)


docs/supported-mutators.md
```diff
@@ -56,13 +56,15 @@ All Stryker versions support a variety of different mutators. We've aligned on a
 | MultiplicationAssignmentNegation              | `*=`                 | `/=`                 |
 | DivisionAssignmentNegation                    | `/=`                 | `*=`                 |
 | RemainderAssignmentToMultiplicationAssignment | `%=`                 | `*=`                 |
-| LeftShiftAssignmentNegation                   | `<<=`                | `>>=`                |
-| RightShiftAssignmentNegation                  | `>>=`                | `<<=`                |
+| LeftShiftAssignmentNegation                   | `<<=`                | `>>=`, `>>>=`²       |
```
rouke-broersma, line 59: Negation can only have one outcome, not two. This should be documented as two distinct mutations

```suggestion
| LeftShiftAssignmentToRightShiftAssignment                   | `<<=`                | `>>=`, `>>>=`²       |
| LeftShiftAssignmentToUnsignedRightShiftAssignment                   | `<<=`                | `>>=`, `>>>=`²       |
```

docs/supported-mutators.md
```diff
@@ -204,6 +206,8 @@ Due to differences in language syntax, method expressions are implemented differ
 | MaxToMin                        | `Max()`               | `Min()`             |
 | SumToCount                      | `Sum()`               | `Count()`           |
 | CountToSum                      | `Count()`             | `Sum()`             |
+| Append                          | `Append()`            | `Prepend()`         |
+| Prepend                         | `Prepend()`           | `Append()`          |
```
rouke-broersma, line 210:
```suggestion
| AppendToPrepend                          | `Append()`            | `Prepend()`         |
| PrependToAppend                         | `Prepend()`           | `Append()`          |
```

Alternatively 

```suggestion
| AppendNegation                          | `Append()`            | `Prepend()`         |
| PrependNegation                         | `Prepend()`           | `Append()`          |
```

However I'm not sure if that change sufficiently natural for usage of Negation. We also did not use Negate for Any and All so I think I prefer the To variant. @nicojs what do you think

rouke-broersma (changes requested):

docs/supported-mutators.md
rouke-broersma: Took another look at the original mutator name proposal and Negation does seem to fit with the agreed upon naming so a breaking change is not worth it. I've reverted that part of the change.

rouke-broersma (approved):

PR #3860, opened 2025-08-12
Python Code Highlighting support

vivganes:
I am working on an adpater for python's mutation testing tools to `mutation-testing-elements` 

During that I realized that we do  not support python code highlighting


hugo-vrijswijk:
Great to have the adapter for python! I should mention that although mutation-testing-elements is (mostly) made by Stryker people, but doesn't have much to do with it specifically. There's no Stryker branding or mentions in the report. It's just a standard for JSON schema

PR #4099, opened 2026-03-13
fix(elements): compatibility with `History` routers

hugo-vrijswijk:
If a router uses the `history.pushState` API, that does not fire the `hashchange` event. Instead, we should listen to the `navigation.onnavigatesuccess` event, which is fired when the active history entry changes. This allows us to support both hash-based routers and history-based routers. Because `navigation` is a newer API, we keep the `hashchange` event listener too


nicojs:
Awesome fix!

